PR #1, opened 2017-01-18
IP-456: New Relic

mezis:
This is work in progress; looking for comments/suggestions.

Adds a simple Railtie that loads and starts New Relic from an initialiser, with a configuration.
The Railtie will abort if the license key is missing, or if the app contains a `new_relic.yml` file.

Tests rely on:

1. the `build_test_app` spec support, which:

    - runs `rails new` to create a dummy app
    - configures the app (for now, only adding the `roo_on_rails` gem and a default `.env` file)
    - caches the app across specs

2. the `run_test_app`spec support, which

    - is shamelessly adapted from [Routemaster's](https://github.com/deliveroo/routemaster/blob/master/spec/support/integration.rb#L10)
    - allows to detect what the app logs, and its exit status

TODO:

- [x] Write tests
- [x] Set up CI
- [x] Set up Code Climate

===


roo_on_rails.gemspec
```diff
@@ -30,7 +30,13 @@ Gem::Specification.new do |spec|
...
 
+  spec.add_runtime_dependency 'dotenv-rails', '~> 2.1'
+  spec.add_runtime_dependency 'newrelic_rpm', '~> 3.17'
+  spec.add_runtime_dependency 'rails', '~> 5.0'
+
   spec.add_development_dependency "bundler", "~> 1.13"
   spec.add_development_dependency "rake", "~> 10.0"
   spec.add_development_dependency "rspec", "~> 3.0"
+  spec.add_development_dependency 'thor', '~> 0.19'
+  spec.add_development_dependency 'pry'
```
tompave-roo: I hate to be that guys... but why not `pry-byebug` then?
Also because using `pry` instead of `irb` for `rails console` is a must.

gregbeech: @tompave-roo you don't really need it, you can just use `byebug` instead of `binding.pry` and then have many more options available.

tompave-roo: I was actually thinking of _literally_ running `rails console` in the terminal.

gregbeech: Yeah, you can do that. `byebug` will work if you do that.

gregbeech: Oh, sorry, you don't mean for debugging? Does pry console do anything much more than IRB other than allowing debugging?

tompave-roo: well, yes, you can print documentation and source of methods for example.
But you can also `cd` into objects and classes and `ls` their instance methods, variables and ancestors, and move down and up as if the object-oriented hierarchy was a Unix file system.

Here there are more examples: http://pryrepl.org/

PR #2, opened 2017-01-20
IP-456: New Relic bugfix

mezis:
This prevents New Relic from loading in test environments, as that breaks horribly.

===


tompave-roo:
Makes sense, LGTM

PR #3, opened 2017-02-08
IP-456: New Relic

mezis:
As discussed in deliveroo/logistics-dashboard#116, this changes the `NEW_RELIC_LICENSE_KEY` detection to only fail in production environments, when no key has been specified.

===


spec/new_relic_spec.rb
```diff
@@ -26,12 +38,30 @@
...
   context 'when NEW_RELIC_LICENSE_KEY is missing' do
     before do
       app_helper.comment_lines app_path.join('.env'), /NEW_RELIC_LICENSE_KEY/
     end
   
-    include_examples 'abort early', /NEW_RELIC_LICENSE_KEY is required/
+    context 'in the test environment' do
+      let(:app_env) { 'test' }
+
+      include_examples 'does not load'
```
gregbeech: Why doesn't it load in test but it does in development? The code looks to me like they should behave the same? Also it feels like it would make sense if they behaved the same.

mezis: Mainly because last I checked, New Relic _really_ didn't like to be loaded in a test environment. But you're right though, it should be consistent; let me see if I can coax it.

gregbeech: Ah OK so it's not specifically related to the code here, just part of what happens in the NR gem itself? In that case I think it's fine to have it different, but maybe add a comment to the tests as to why it's different.

Do we actually want/need it to load in dev anyway though?

mezis: So, I checked again, and when New Relic gets loaded in `test`, this sort of thing happens to your test suite (from `logistics-dashboard`):

```
.** [NewRelic][2017-02-08 17:41:09 +0100 salsa.local (35030)] ERROR : Error during start callback for event 'process_action.action_controller':
** [NewRelic][2017-02-08 17:41:09 +0100 salsa.local (35030)] ERROR : NameError: uninitialized constant AnonymousController
** [NewRelic][2017-02-08 17:41:09 +0100 salsa.local (35030)] ERROR : Debugging backtrace:
....rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/newrelic_rpm-3.17.2.327/lib/new_relic/agent/instrumentation/action_controller_subscriber.rb:78:in `const_get'
.../logistics-dashboard/spec/controllers/concerns/oauth_session_required_spec.rb:14:in `block (2 levels) in <top (required)>'
```

ie. the agent gets really confused with the anonymous/dynamic stuff going on in a test suite.

gregbeech: Yep probably don't want that happening in test.

PR #4, opened 2017-02-08
IP-460: Heroku-friendly configuration

mezis:
This is a preliminary version of the `roo_on_rails` script, which checks if your app is properly set up(and fixes it for you where possible).

Example output so far:

### Features

Running `roo_on_rails` will check if you app exists on Heorku with the right names and has preboot enabled.

### Under the hood

- `exe/roo_on_rails` just calls `RooOnRails::Harness`.
- `Harness` just runs a number of `Checks` in sequence
- each `Check` has a `call` method to perform the check; can have a `fix` method to attempt a fix; and may update some `context` for other checks.
- each `Check` can `require` other checks, forming a dependency tree.

### Left to do

- [x] Tests. Any input on how to best do this is welcome.
- [x] Dealing with nonstandard app naming (i.e. different from `{repo}-{env}`).
- [ ] Usage documentation (in README).
- [ ] Document how to write checks.

===

Jira story [#IP-460](https://deliveroo.atlassian.net/browse/IP-460) in project *Infrastructure and Platform*:

> Check/set Heroku-friendly configuration
> 
> - dotenv integration
> - preboot, runtime-metrics, any other app setup
> - buildpacks
> - dd bridge config
> - app generation, git remote setup, addons, configure auto-deploys


lib/roo_on_rails/checks/base.rb
```diff
@@ -0,0 +1,72 @@
...
+      end
+
+      def _run(cmd)
+        result = Bundler.with_clean_env { %x{#{cmd}} }
+        return [$?.success?, result]
+      end
+
+      def _run!(cmd)
+        Bundler.with_clean_env { system(cmd) }
+        raise CommandFailed.new(cmd) unless $?.success?
```
gregbeech: Might be nice to have the result as part of the error message here, e.g.

```ruby
succeeded, result = _run(cmd)
raise CommandFailed.new(cmd, succeeded) unless succeeded
```

lib/roo_on_rails/checks/git/origin.rb
```diff
@@ -0,0 +1,24 @@
...
+      class Origin < Base
+        def _intro
+          "Checking your Git origin remote..."
+        end
+
+        def _call
+          status, url = _run "git config remote.origin.url"
+          _fail "Origin does not seem to be configured." unless status
+          
+          org, repo = url.strip.gsub(%r{\.git$}, '').split(%r{[:/]}).last(2)
```
gregbeech: Could just use `sub` here as it's anchored to the end. Also I believe it's considered better to use `\Z` rather than `$` in Ruby unless you specifically mean the end of a line in a multiline string?

lib/roo_on_rails/checks/base.rb
```diff
@@ -0,0 +1,72 @@
...
+        raise unless @fix
+        say "\t· attempting to fix", %i[yellow]
+        _fix
+        say "\t· re-checking", %i[yellow]
+        _call
+      end
+
+      protected
+
+      def _intro
```
gregbeech: As you know I'm generally not keen on the leading underscore in Ruby as it doesn't seem to be a common convention. I can just about live with it for private methods, but it seems egregious when used for methods that are _intended_ to be overridden by other classes.

lib/roo_on_rails/checks/heroku/app_exists.rb
```diff
@@ -0,0 +1,44 @@
...
+          end
+
+          _state.heroku.app![@_env] = correct_app.first
+          _ok "found app #{bold correct_app.first}"
+        end
+
+        private
+
+        def _client
+          _state.heroku.api_client
```
gregbeech: Where does this `_state.heroku` come from? I can't see it being set anywhere (or state, for that matter). What am I missing?

lib/roo_on_rails/checks/multi.rb
```diff
@@ -0,0 +1,17 @@
+module RooOnRails
+  module Checks
+    class Multi
+      def initialize(variants:, of:)
+        @_variants = variants
+        @_of = of
```
gregbeech: I don't really see the point of `@_` here, especially as this class isn't designed to be mixed, inherited, etc. What's wrong with just `@`? IMHO this convention makes the code needlessly harder to read.

gregbeech:
> Dealing with nonstandard app naming (i.e. different from {repo}-{env}).

We know there are a couple of conventions already in use:

* `deliveroo-{repo}-{env}`
* `roo-{repo}-{env}`

So could also check those (in fact, I'd probably be inclined to check those first as they're more specific). I guess it would make sense to have a `.roo_on_rails.yml` file where we can read/write this kind of setting, and allow the user to input the pre-{env} part if we can't detect the app.

lib/roo_on_rails/checks/base.rb
mezis: sure.

lib/roo_on_rails/checks/git/origin.rb
mezis: If `git config remote.origin.url` doesn't return a single-line URL we have bigger problems. I think using `\Z` everywhere is a bit pedantic when the input is well-known ;)
(point taken on `sub`, will fix)

lib/roo_on_rails/checks/multi.rb
mezis: "harder to read" is way debatable, but sure. Why not.

lib/roo_on_rails/checks/heroku/app_exists.rb
mezis: `_state` is defined in the base class. It's passed on by the constructor (ie. from the harness); it's a `Hashie::Mash`.

The gist as that (as tentatively explained above): checks can depend on the output of other checks, and they communicate through the `_state` because it'd be very inconvenient to have checks _call_ eachother.

Not the most elegant but it's flexible. [Interactor](https://github.com/collectiveidea/interactor#interactor) does this, for instance. I could call it "context" instead of "state" if that's more clear?

lib/roo_on_rails/checks/base.rb
mezis: The status code? Yeah, possibly. Although I haven't needed it yet.
Also, `_run!` can't use `_run` because the former calls to `system` (doesn't capture STDOUT) and the latter calls backtick (does capture).

mezis:
(Copying this comment at top level for discussion)

`_state`, now renamed to `context`, is defined in the base class. It's passed on by the constructor (ie. from the harness); it's a `Hashie::Mash`.

The gist as that (as tentatively explained above): checks can depend on the output of other checks, and they communicate through the `_state` because it'd be very inconvenient to have checks _call_ eachother.

Not the most elegant but it's flexible. [Interactor](https://github.com/collectiveidea/interactor#interactor) does this, for instance.
Thoughts?

lib/roo_on_rails/checks/base.rb
```diff
@@ -0,0 +1,100 @@
...
+      def pass(msg)
+        say "\t✔ #{msg}", :green
+      end
+
+      def fail!(msg)
+        say "\t✘ #{msg}", :red
+        raise Failure, msg
+      end
+
+      def fail‼︎(msg)
```
gregbeech: Is that a unicode double-exclamation mark character? I'd be inclined to keep things in the range of stuff I can type, either `def fail!(msg, final: false)` or `def fail_final!(msg)`.

lib/roo_on_rails/checks/base.rb
```diff
@@ -0,0 +1,100 @@
...
+      def resolve(deps)
+        deps.each do |dep|
+          context.deps ||= {}
+          context.deps[dep.signature] ||= 
+            begin
+              # sig = signature.join('/')
+              # dep_sig = dep.signature.join('/')
+              # say "┌ resolving #{sig} → #{dep_sig}"
+              dep.run
+              # say "└ resolved  #{sig} → #{dep_sig}"
```
gregbeech: What's the purpose of the commented out stuff? Debug logging? If so, might be nice to have it actually outputtable as debug logging with some kind of log level setting.

lib/roo_on_rails/checks/heroku/app_exists.rb
```diff
@@ -0,0 +1,69 @@
...
+
+        def intro
+          "Checking if #{bold env} app exist..."
+        end
+
+        def call
+          all_apps = client.app.list.map { |a| a['name'] }
+          matches = all_apps.select { |a| candidates.include?(a) }
+
+          unless matches.one?
```
gregbeech: Isn't this effectively the same condition as `if matches.many?` below?

lib/roo_on_rails/checks/heroku/app_exists.rb
```diff
@@ -0,0 +1,69 @@
...
+          end
+
+          context.heroku.app![env] = matches.first
+          pass "found app #{bold matches.first}"
+        end
+
+        private
+
+        def name_stem
+          context.app_name_stem || context.git_repo.gsub('.', '')
```
gregbeech: CodeClimate suggests using `delete` instead of `gsub`.

lib/roo_on_rails/checks/heroku/app_exists.rb
```diff
@@ -0,0 +1,69 @@
...
+        end
+
+        def candidates
+          [
+            [nil, 'roo', 'deliveroo'],
```
gregbeech: I'd reverse the order of these so we search for the more specific `deliveroo-` and `roo-` first, to reduce the chance of us detecting somebody else's app.

lib/roo_on_rails/shell.rb
```diff
@@ -0,0 +1,23 @@
...
+  class Shell
+    CommandFailed = Class.new(StandardError)
+
+    def run(cmd)
+      result = Bundler.with_clean_env { %x{#{cmd}} }
+      return [$?.success?, result]
+    end
+
+    def run!(cmd)
+      success, _ = run(cmd)
```
gregbeech: You're not using `success` here... did you mean to?

lib/roo_on_rails/shell.rb
```diff
@@ -0,0 +1,23 @@
...
+      raise CommandFailed.new(cmd) unless $?.success?
+    end
+
+    def run?(cmd)
+      success, _ = run(cmd)
```
gregbeech: You're not using `success` here... did you mean to?

spec/roo_on_rails/checks/git/origin_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+describe RooOnRails::Checks::Git::Origin do
+  let(:shell) { FakeShell.new }
+  let(:context) { Hashie::Mash.new }
+
+  subject { described_class.new(shell: shell, context: context) }
+
+  let(:perform) { silence_stream(STDOUT) { subject.run } }
+
+
```
gregbeech: Being ultra-picky... two blank lines 😢 🐼

lib/roo_on_rails/checks/base.rb
mezis: I intended this to be a prompt for naming suggestions. Thanks :)

lib/roo_on_rails/checks/heroku/app_exists.rb
mezis: It is. Leftover output.

lib/roo_on_rails/checks/base.rb
mezis: Yep, this is debug logging. I'll strip for now as the priority is shipping this first increment.

lib/roo_on_rails/shell.rb
mezis: Nope. Fixing!

gregbeech:
I realise there's more to do on your checklist but I think this is a solid base and would be worth getting in once you're happy it's been reviewed by enough people.

PR #6, opened 2017-03-06
Wider build matrix

mezis:
Tests across multiple Rubies and Rails versions.

Rationale:
- `roo_on_rails` adoption will be easier if we don't force apps to upgrade to Rails 5 on Ruby 2.3 first.
- compiling on `ruby-head` can give us some advance notice if things start failing.


.travis.yml
```diff
@@ -1,8 +1,14 @@
 sudo: false
 language: ruby
 rvm:
-  - 2.3.3
+  - 2.2.5
+  - 2.3.1
```
martyphee: Should this be `2.3.3` or is the min version?

.travis.yml
```diff
@@ -12,3 +18,8 @@ cache:
 before_script:
   - unset RACK_ENV
   - unset RAILS_ENV
+matrix:
+  allow_failures:
+    - rvm: ruby-head
+    - rvm: 2.4.0
```
jphastings: :+1:

.travis.yml
mezis: Using `2.3.1` here because Travis machines used to have a few Rubies pre-baked until recently, which made for faster builds.
[Turns out](http://rubies.travis-ci.org/) they only boot with 1.9.3 and install the right one now, so I'll change to use the latest patchlevel of each Ruby version.

PR #7, opened 2017-03-06
Ip 471/ssl timeouts deflate

danielcooper:
This pr:
- Inserts rack timout into the middleware chain
- Inserts a `ActiveRecord::Base.connection` clearer (taken from LD)
- sets ssl to true


lib/roo_on_rails/railties/http.rb
```diff
@@ -0,0 +1,31 @@
...
+      initializer 'roo_on_rails.http' do |app|
+        $stderr.puts 'initializer roo_on_rails.http'
+        require "rack/timeout/base"
+        require "roo_on_rails/rack/safe_timeouts"
+
+        ::Rack::Timeout.service_timeout = ENV.fetch('RACK_SERVICE_TIMEOUT', 15).to_i
+        ::Rack::Timeout.wait_timeout = ENV.fetch('RACK_WAIT_TIMEOUT', 30).to_i
+        ::Rack::Timeout::Logger.level = Logger::WARN
+
+        Rails.application.config.middleware.insert_before(
```
mezis: Any reason you're using `Rails.application` here instead of `app`?

lib/roo_on_rails/railties/http.rb
```diff
@@ -0,0 +1,31 @@
...
+        # This needs to be inserted low in the stack, before Rails returns the
+        # thread-current connection to the pool.
+        app.config.middleware.insert_before(
+          ActionDispatch::Cookies,
+          RooOnRails::Rack::SafeTimeouts
+        )
+
+        app.config.middleware.use ::Rack::Deflater
+
+        app.config.use_ssl = true unless Rails.env.test?
```
mezis: I've seen `use_ssl` before but I can't find it documented anywhere.
I think [`force_ssl`](http://blog.bigbinary.com/2016/08/24/rails-5-adds-more-control-to-fine-tuning-ssl-usage.html) is what you're aiming fore here, though.

I tend to use [`rack-ssl-enforcer`](https://github.com/tobmatth/rack-ssl-enforcer) as it's well-tested, documented, not Rails-dependent and clear on what it does, but up to you.

Either way - this isn't test-covered; sorry I missed that during the first review.

danielcooper: Sure. The middleware will give me a thing to test (no idea how to test if the config is set).

lib/roo_on_rails/railties/new_relic.rb
```diff
@@ -0,0 +1,29 @@
...
+
+        if %w[ test development ].exclude?(Rails.env.to_s) and license_key == 'override-me'
+          abort 'Aborting: NEW_RELIC_LICENSE_KEY must be set in production environments'
+        end
+
+        if license_key.nil?
+          abort 'Aborting: NEW_RELIC_LICENSE_KEY is required'
+        end
+
+        path = %w[newrelic.yml config/newrelic.yml].map { |p|
```
nronas: Just a personal preference suggestion here, i would make the list of pathnames into a constant and then just find which one exists, like: 

```ruby
SUPPORTED_PATHS = [
  Pathname.new('newrelic.yml'),
  Pathname.new('config/newrelic.yml')
].freeze

path = SUPPORTED_PATHS.find(&:exist?)
```

mezis: This was existing code - Dan just moved it around, I believe.
Personally, I would not extract this into a constant unless it's reused, as that lowers readability.

mezis:
So, I'm being a pain, but here's extra requirements for you @danielcooper:

1. We need to be able to disable these 3 middleware. Sample use case is that (apparently) iOS doesn't like `Rack::Deflate` too much. I'd suggest `ROO_ON_RAILS_NO_RACK_DEFLATE=1`, etc.
2. Mention the changes in the README?

If you'd rather merge and do this later / in another ticket that's fine with me.

danielcooper:
@mezis No, that's fine by me. Will do shortly

mezis:
Thanks man. Cutting a release.

PR #8, opened 2017-03-09
IP-457: GitHub branch protection

gregbeech:
This change introduces a GitHub branch protection check for CI plus code analysis and coverage, as well as a couple of other things around code reviews and push restrictions. It also adds a task to obtain a GitHub token as a necessary prerequisite.

I did look at splitting this up into multiple tasks (i.e. one for CI, one for reviews, one for branch protection) but because the update request is a PUT rather than a PATCH you end up having to reflect back all the original configuration or you wipe it out. The end result was that there was a lot more code but it actually felt less maintainable, so I collapsed it back into a single check and refactored that for readability. I think this is the right compromise.

--

❓  I haven't written much in the way of tests for this. I can do, of course, but essentially the implementation of them all will be mocking methods on `Octokit::Client` that I'm calling and then checking the code calls then in the sequence I expect which, of course, it will because that's how I wrote it. As such, I'm somewhat unsure of the value of doing this, because really it's more that the code I've manually tested to work would be testing that I've written the mocks correctly, than that the mocks are testing code I already know works because I manually tested it. Thoughts?

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,29 @@
+codecov:
+  notify:
+    require_ci_to_pass: yes
+
+coverage:
+  precision: 2
+  round: down
+  range: "70...100"
+
+  status:
+    project:
+      default:
+        enabled: yes
+        threshold: 1%
+    patch:
+      default:
+        enabled: yes
+        threshold: 5%
+    changes: no
+
+parsers:
+  gcov:
+    branch_detection:
+      conditional: yes
+      loop: yes
+      method: no
+      macro: no
+
+comment: off
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,32 @@
+AllCops:
+  TargetRubyVersion: 2.4
+
+Style/FrozenStringLiteralComment:
+  Enabled: false
+
+Style/ModuleFunction:
+  EnforcedStyle: extend_self
+
+# Disable Metric checks for specs
+Metrics/BlockLength:
+  Exclude:
+    - 'spec/**/*.rb'
+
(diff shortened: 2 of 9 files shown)


mezis:
Re. testing:

- one direction could be recording Github API responses and Webmocking /VCR'ing the hell out of it. I'm not sure it'd add huge value either.
- another would be to integration-test this, which would involves creating (then tearing down) repos. That'd probably be more valuable, but more painful.

I'm on the fence: I wouldn't want `roo_on_rails` (a foundational tool) to give a bad example of poor test coverage, or to be unstable. But OTOH I'd also like it to be shipped fairly rapidly and used in all apps...

gregbeech:
> one direction could be recording Github API responses and Webmocking /VCR'ing the hell out of it. I'm not sure it'd add huge value either.

This feels like a slightly more painful way to check that the code I know works, works... if you changed the implementation then you'd probably just re-record the VCR stuff so it wouldn't provide much/any protection for refactoring.

> another would be to integration-test this, which would involves creating (then tearing down) repos. That'd probably be more valuable, but more painful.

Yeah, could do that... trouble is that the only way to check that the code has done the right thing is to, erm, copy and paste the implementation of itself into the tests because it's already the bit of code that checks that a repo is in the right state!

> I'm on the fence: I wouldn't want roo_on_rails (a foundational tool) to give a bad example of poor test coverage, or to be unstable

Agreed. At the same time we also don't want it to be an example of adding tests without value, where the implementation is effectively testing the tests, or where the tests are simply a copy of the implementation. I'm struggling to find a way to write tests that I feel adds value.

PR #10, opened 2017-03-21
IP-469: Set database timeout to 200ms by default

gregbeech:
This change automatically set the DB statement timeout to 200ms unless specifically overridden by an environment variable.

This changes the database used by the app to Postgres to the timeout can be queried, and adds a rake task to display it as I couldn't see any other way of executing it in the context of the Rails app (thanks @danielcooper for the idea).


lib/roo_on_rails/railties/database.rb
```diff
@@ -0,0 +1,12 @@
+module RooOnRails
+  module Railties
+    class Database < Rails::Railtie
+      initializer 'roo_on_rails.database', after: 'active_record.initialize_database' do |_app|
+        $stderr.puts 'initializer roo_on_rails.database'
+
+        ActiveRecord::Base.configurations[Rails.env]['timeout'] = ENV.fetch('DATABASE_TIMEOUT', 200)
```
mezis: Does this actually set the statement timeout for Postgres? Or the connection timeout?

gregbeech: Bugger, I've been looking at entirely the wrong timeout... what an idiot.

gregbeech: @mezis fortunately the approach was sound, it now sets the statement timeout.

gregbeech:
OK so this does work, I've tested it by using the local gem in logistics dashboard and run a `SHOW statement_timeout` command on the ActiveRecord connection. But given sqlite doesn't have the concept of statement timeouts there's no way I can see to test that it's working, unless we use something like postgres as the default database in the test app -- but do we really want to go there?

mezis:
> unless we use something like postgres as the default database in the test app -- but do we really want to go there?

Why not? All the Rails versions we're using can pick up a `DATABASE_URL`, I think*; and Postgres is already on your machine and easy to install in CI?

* if they don't that's maybe a good ticket to add to `roo_on_rails`!

gregbeech:
> Postgres is already on your machine and easy to install in CI?

gregbeech:
See #13 instead.

PR #11, opened 2017-03-21
IP-470/Sidekiq and Hirefire setup

danielcooper:
This PR:

1. Prompts the user to install sidekiq. Not sure if there's a way to get sidekiq pro into the build without some user action
2. Sets up easy hirefire scaling based on queue SLAs

Todo:
- [x] Default Queue names
- [x] Sidekiq.yml

Should we just generate a sidekiq yml with sensible defaults. When the user wants to schedule jobs they'll be needing one anyway and it'll simplify the code here (no need to have more config vars)?


README.md
```diff
@@ -95,6 +95,21 @@ We'll insert the following middlewares into the rails stack:
 2. `Rack::SslEnforcer`: enforces HTTPS.
 3. `Rack::Deflater`: compresses responses from the application, can be disabled with `ROO_ON_RAILS_RACK_DEFLATE` (default: 'YES').
 
+### Sidekiq
+
+Unless `SIDEKIQ_ENABLED` is set to false we'll check for the existance of sidekiq gems. Please ensure the following lines are included in your project Gemfile.
+
+```
+gem 'sidekiq-scheduler'
```
gregbeech: Should probably have this in a `source 'https://rubygems.org'` block too?

lib/roo_on_rails/checks/sidekiq/sidekiq.rb
```diff
@@ -0,0 +1,43 @@
...
+    module Sidekiq
+      class Sidekiq < Base
+        WORKER_PROCFILE_LINE = 'worker: RAILS_MAX_THREADS=$SIDEKIQ_DB_POOL_SIZE DB_REAPING_FREQUENCY=$SIDEKIQ_DB_REAPING_FREQUENCY bundle exec sidekiq'.freeze
+
+        def intro
+          'Checking Sidekiq Setup...'
+        end
+
+        def call
+          unless ENV.fetch('SIDEKIQ_ENBALED', 'true').to_s.match?(/\A(YES|TRUE|ON|1)\Z/i)
```
gregbeech: "SIDEKIQ_ENABLED" (and the line below)

lib/roo_on_rails/checks/sidekiq/sidekiq.rb
```diff
@@ -0,0 +1,43 @@
...
+            fail! 'Sidekiq is not installed, see the README: https://github.com/deliveroo/roo_on_rails#sidekiq'
+          end
+        end
+
+        def check_for_procfile
+          return if File.exist?('Procfile') && File.read('Procfile').include?('worker')
+          say 'No Procfile with workers entry found. Adding one'
+          if File.exist?('Procfile')
+            raise 'appending'
+            append_to_file 'Procfile', "\n#{WORKER_PROCFILE_LINE}"
```
gregbeech: This should be in `fix` rather than `call`.

lib/roo_on_rails/default.env
```diff
@@ -9,3 +9,8 @@ NEW_RELIC_DEVELOPER_MODE=false
 NEW_RELIC_LICENSE_KEY=override-me
 RACK_SERVICE_TIMEOUT=10
 RACK_WAIT_TIMEOUT=30
+SIDEKIQ_ENABLED=true
+SIDEKIQ_THREADS=25
+SIDEKIQ_DB_POOL_SIZE=26
```
gregbeech: This is the recommendation (thread count + 1) but it means we definitely need pg_bouncer set up by roo_on_rails too!

lib/roo_on_rails/railties/sidekiq.rb
```diff
@@ -0,0 +1,23 @@
+module RooOnRails
+  module Railties
+    class Sidekiq < Rails::Railtie
+      initializer 'roo_on_rails.sidekiq' do |app|
+        require 'hirefire-resource'
+        $stderr.puts 'initializer roo_on_rails.sidekiq'
+        break unless ENV.fetch('SIDEKIQ_ENBALED', 'true').to_s =~ /\A(YES|TRUE|ON|1)\Z/i
```
gregbeech: "SIDEKIQ_ENABLED"

lib/roo_on_rails/sidekiq_sla_metric.rb
```diff
@@ -0,0 +1,73 @@
+# Returns stats on the current SLA performance of queues in a sidekiq instance.
+# Assumes workers are not bound to queues
+require 'sidekiq/api'
+require 'active_support'
+require 'active_support/core_ext/numeric'
+module RooOnRails
+  module SidekiqSlaMetric
```
gregbeech: Would probably be a bit nicer for this to be `RooOnRails::Sidekiq::SlaMetric`

lib/roo_on_rails/railties/sidekiq.rb
```diff
@@ -0,0 +1,23 @@
...
+        config_hirefire(app)
+      end
+
+      def config_hirefire(app)
+        return unless ENV['HIREFIRE_TOKEN']
```
gregbeech: Maybe should add a warning that autoscaling isn't enabled if this isn't set?

lib/roo_on_rails/default.env
mezis: Wooooo that's a high thread count. The assumption you're making is that the average job is less than 4% CPU or more than 96% I/O. Definitely not the norm except for apps that mostly use jobs to call into APIs...

A default of 5 threads would make much more sense to me.

README.md
```diff
@@ -95,6 +95,28 @@ We'll insert the following middlewares into the rails stack:
...
+source 'https://rubygems.org' do
+  gem 'sidekiq-scheduler'
+  gem 'sidekiq-unique-jobs'
+end
+source 'https://gems.contribsys.com/' do
+  gem 'sidekiq-pro'
+end
+```
+
+Sidekiq pro requires some authentication to be installed. Please get the key from your Tech Lead (or password manager of choice).
```
mezis: This shouldn't be necessary is gems are cached in the repo (which we already decided they should, particularly for these 3rd party gems)?

README.md
```diff
@@ -95,6 +95,28 @@ We'll insert the following middlewares into the rails stack:
 2. `Rack::SslEnforcer`: enforces HTTPS.
 3. `Rack::Deflater`: compresses responses from the application, can be disabled with `ROO_ON_RAILS_RACK_DEFLATE` (default: 'YES').
 
+### Sidekiq
+
+Unless `SIDEKIQ_ENABLED` is set to false we'll check for the existance of sidekiq gems. Please ensure the following lines are included in your project Gemfile.
+
+```
+source 'https://rubygems.org' do
```
mezis: Why not just depend on these gems? I can't think of any app that won't use _any_ jobs, and worst case it'll carry a couple of extra gems.

lib/roo_on_rails/checks/sidekiq/settings.rb
```diff
@@ -0,0 +1,29 @@
...
+    module Sidekiq
+      class Settings < Base
+        requires Sidekiq
+        def intro
+          'Checking Sidekiq settings.yml...'
+        end
+
+        def call
+          fail! 'No sidekiq settings found' unless File.exist?('config/sidekiq.yml')
+          return if File.read('config/sidekiq.yml') == yaml_template
```
mezis: Do we really need this Yaml file? Sidekiq can't take settings programmatically?

danielcooper: Not that I can find. Happy to change if you know how.

README.md
danielcooper: It's just because of the sidekiq-pro gem. I don't see how we can bundle it with this gem so it needs to be added by hand.

mezis: Well, see my next comment — if we make caching gems (`bundle package --all`) the norm, then you can just have `roo_on_rails` depend on `sidekiq-pro`, and not have to bother with these comments.

(and possibly issue a nice warning if it's not available)

lib/roo_on_rails/checks/sidekiq/sidekiq.rb
```diff
@@ -0,0 +1,48 @@
+require 'roo_on_rails/checks/base'
+require 'thor'
+
+module RooOnRails
+  module Checks
+    module Sidekiq
+      class Sidekiq < Base
+        WORKER_PROCFILE_LINE = "worker: RAILS_MAX_THREADS=$SIDEKIQ_DB_POOL_SIZE \
+                                DB_REAPING_FREQUENCY=$SIDEKIQ_DB_REAPING_FREQUENCY \
```
mezis: What impact does `DB_REAPING_FREQUENCY` have? It's not documented in Sidekiq or in Rails (that I can find).

lib/roo_on_rails/checks/sidekiq/settings.rb
mezis:
```ruby
# in your initializer
Sidekiq.configure_server do |x|
  x.options[:concurrency] = ENV.fetch('SIDEKIQ_THREADS')
  x.options[:queues] = %w[realtime ... within1day]
end
```

This could let you get rid of the generated config file and keep your Procfile line down to `bundle exec sidekiq`.

lib/roo_on_rails/default.env
```diff
@@ -9,3 +9,8 @@ NEW_RELIC_DEVELOPER_MODE=false
 NEW_RELIC_LICENSE_KEY=override-me
 RACK_SERVICE_TIMEOUT=10
 RACK_WAIT_TIMEOUT=30
+SIDEKIQ_ENABLED=true
+SIDEKIQ_THREADS=5
+SIDEKIQ_DB_POOL_SIZE=6
+SIDEKIQ_DB_REAPING_FREQUENCY=10
```
mezis: - These new settings aren't documented in the README.
- It feels like `SIDEKIQ_DB_POOL_SIZE` should always be set to `SIDEKIQ_THREADS` + 1 (ie. not be configurable).
- OTOH it looks like `SIDEKIQ_DB_POOL_SIZE` isn't used anywhere( guessing it needs to be hooked up to the database configuration somehow?
- Same for `SIDEKIQ_DB_REAPING_FREQUENCY` as mentioned in another comment?

danielcooper: I've juyst taken these directly from LD as a starting point.

Looking into it, `SIDEKIQ_DB_REAPING_FREQUENCY` indirectly sets `DB_REAPING_FREQUENCY` which  in tern sets ActiveRecords reaping_frequency. With large amounts of sidekiq concurrency, as I understand it, you'd want a different value for workers vs the web server.

I'm going to park this ticket until https://github.com/deliveroo/roo_on_rails/pull/10 lands. While we probably don't want to change the default settings now it's important they're configurable and that ticket provides the pluming for setting it up

gregbeech: `SIDEKIQ_THREADS` should really be 25 (or not set) as that's Sidekiq's normal default.

Ideally `SIDEKIQ_DB_POOL_SIZE` should be `SIDEKIQ_THREADS + 1` -- it's set in the `Procfile` so maybe we can do maths in there? Does something like `$(SIDEKIQ_THREADS + 1)` work in those?

`SIDEKIQ_DB_REAPING_FREQUENCY` sets `DB_REAPING_FREQUENCY` in the `Procfile` as @danielcooper says. The reason we don't set it as a global thing is that it's not recommended to be used for web processes, only for background processes.

gregbeech: Note though, if we're setting the Sidekiq connection size to that number, then pg_bouncer becomes a prerequisite so we'd want/need to add that to RoR pretty sharpish.

lib/roo_on_rails/checks/sidekiq/sidekiq.rb
gregbeech: You need this in `database.yml`:

```yaml
  reaping_frequency: <%= ENV['DB_REAPING_FREQUENCY'] %>
```

README.md
gregbeech: Did we decide that? Any links to the guidelines that we wrote on deliveroo.engineering 😉 ?

lib/roo_on_rails/checks/sidekiq/sidekiq.rb
mezis: That's not documented anywhere though. Shouldn't this maybe be enforced by the gem if we're going to rely on it?

lib/roo_on_rails/default.env
mezis: Or we could reduce the default. I'm quite unconvinced that 25 threads makes sense for the majority of use cases; but either way, a low default might protect us a bit until we do have `pg_bouncer` set up everywhere.

gregbeech: We're actually running at 40 threads because of all the IO; it reduced our dyno count by around 50% compared to the previous week. So I'm not sure there is a sensible default. Was just suggesting that if you don't set anything at all Sidekiq will use 25 and I expect most people would assume that's the case.

lib/roo_on_rails/checks/sidekiq/sidekiq.rb
gregbeech: I'm assuming that's in the configuration that @danielcooper is generating?

gregbeech: Ah, no, he's generating the Sidekiq configuration. In which case yes, it should be.

danielcooper:
@gsb, @mezis 

As per your comments:

- SIDEKIQ_THREADS defaults to 25 with the db value set to +1 with bash math (is there a better way to do this?)
- the yml file is no more
- Added the reaper setting to the db

mezis:
> `SIDEKIQ_THREADS defaults to 25 with the db value set to +1 with bash math (is there a better way to do this?)`

I'm pretty skeptical about the `bash` math as it breaks for me locally :)

How about inserting this in the Procfile instead?

```
worker: roo_on_rails sidekiq
```

And have `roo_on_rails` load bundler, then set the appropriate vars and start Sidekiq with the appropriate flags.

lib/roo_on_rails/checks/sidekiq/sidekiq.rb
```diff
@@ -5,9 +5,7 @@ module RooOnRails
   module Checks
     module Sidekiq
       class Sidekiq < Base
-        WORKER_PROCFILE_LINE = "worker: RAILS_MAX_THREADS=($SIDEKIQ_THREADS + 1) \
-                                DB_REAPING_FREQUENCY=$SIDEKIQ_DB_REAPING_FREQUENCY \
-                                bundle exec sidekiq".freeze
+        WORKER_PROCFILE_LINE = 'worker: bundle exec roo_on_rails sidekiq'.freeze
```
mezis: LGTM, but now you need to test it?

danielcooper:
@mezis We've got a boot and shutdown test for sidekiq. Not sure how (or if we'd want to) go beyond that?

lib/roo_on_rails/checks/sidekiq/sidekiq.rb
```diff
@@ -0,0 +1,45 @@
...
+
+        def fix
+          if File.exist?('Procfile')
+            append_to_file 'Procfile', "\n#{WORKER_PROCFILE_LINE}"
+          else
+            create_file 'Procfile', WORKER_PROCFILE_LINE
+          end
+        end
+
+        def check_for_sidekiq
```
gregbeech: Doesn't look like this method is used.

lib/roo_on_rails/checks/sidekiq/sidekiq.rb
```diff
@@ -0,0 +1,45 @@
+require 'roo_on_rails/checks/base'
+require 'thor'
+
+module RooOnRails
+  module Checks
+    module Sidekiq
+      class Sidekiq < Base
```
gregbeech: Seems a bit weird having the same class name as the containing module (both `Sidekiq`). Maybe it should be called `Procfile` as that seems to be what it's doing?

lib/roo_on_rails/railties/database.rb
```diff
@@ -8,6 +8,7 @@ class Database < Rails::Railtie
           config = ActiveRecord::Base.configurations[Rails.env]
           config['variables'] ||= {}
           config['variables']['statement_timeout'] = ENV.fetch('DATABASE_STATEMENT_TIMEOUT', 200)
+          config['reaping_frequency'] = ENV['DB_REAPING_FREQUENCY']
```
gregbeech: Maybe we should make this `DATABASE_REAPING_FREQUENCY` for consistency with the `DATABASE_STATEMENT_TIMEOUT` variable? Not a huge deal.

lib/roo_on_rails/sidekiq/loader.rb
```diff
@@ -0,0 +1,13 @@
+require 'roo_on_rails/environment'
+module RooOnRails
+  module Sidekiq
+    class Loader
+      def self.run
+        RooOnRails::Environment.load
+        ENV['RAILS_MAX_THREADS'] = (ENV.fetch('SIDEKIQ_THREADS').to_i + 1).to_s
+        ENV['DB_REAPING_FREQUENCY'] = ENV['SIDEKIQ_DB_REAPING_FREQUENCY']
+        exec 'sidekiq'
```
gregbeech: I think this is good enough for this PR, but we're going to need to change this to use pg_bouncer before we'd actually be able to use it as we'd run out connections _very_ quickly otherwise. But this PR is already big enough so lets do that separately.

PR #13, opened 2017-03-23
IP-469: Database configuration 

mezis:
An alternative ending for #10, which drops Rails 3 support for this features.

===

Jira story [#IP-469](https://deliveroo.atlassian.net/browse/IP-469) in project *Infrastructure and Platform*:

> Postgresql should be configured with a per-session `statement_timeout`, defaulting to 200ms.
> The `roo_on_rails` script should check this; and the ideally should have a Railtie that makes it the default.
> 
> The simplest option to check this is possibly issuing a `SHOW statement_timeout` request.
> 
> Setting the value from the Railtie might involve overriding default options for the AR database adapter (no idea how).


spec/integration/database_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+
+describe 'Database setup', rails_min_version: 4 do
+  run_test_app
+  let(:app_options) {{ database: 'postgresql' }}
+
+  before { app.wait_start }
+
+  context 'When booting' do
+    before { app_helper.shell_run "cd #{app_path} && rake db:create" }
+    after  { app_helper.shell_run "cd #{app_path} && rake db:drop" }
```
gregbeech: Why not `rake_task` in the above lines?

spec/support/build_test_app.rb
```diff
@@ -6,52 +6,58 @@ module ROR
...
+    RAILS_NEW_BASE_OPTIONS = '--skip-test --skip-git --skip-spring --skip-bundle'
 
     class Helper < Thor::Group
       include Thor::Actions
 
+      def initialize(database: 'sqlite3', keep_scaffold: false)
+        super()
+
+        @database = database
+        @keep_scaffold = false
```
gregbeech: `= keep_scaffold` ?

spec/integration/database_spec.rb
gregbeech: Ah, you removed it. Why's that? Too specific?

README.md
```diff
@@ -95,6 +94,10 @@ We'll insert the following middlewares into the rails stack:
 2. `Rack::SslEnforcer`: enforces HTTPS.
 3. `Rack::Deflater`: compresses responses from the application, can be disabled with `ROO_ON_RAILS_RACK_DEFLATE` (default: 'YES').
 
+### Database configuration
+
+The database statement timeout will be set to a low value by default. Use `DATABASE_STATEMENT_TIMEOUT` (milliseconds, default 200) to customise.
```
gregbeech: This is missing the note I added about it not doing anything in Rails 3. Should probably keep that.

lib/roo_on_rails/railties/database.rb
```diff
@@ -0,0 +1,17 @@
+module RooOnRails
+  module Railties
+    class Database < Rails::Railtie
+      if ActiveRecord::VERSION::MAJOR >= 4
```
gregbeech: I'm not sure this is really needed; setting the variables doesn't _hurt_, it just doesn't do anything.

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# HEAD
+
+Features:
+
+- Sets database statement timeout to 200ms by default (#10).
```
gregbeech: Should be #13 if this is the one we merge.

gregbeech:
Mostly looks reasonable to me. The database workaround feels a bit nasty but I guess if it works then it's OK.

I won't review as I wrote most of it so that would be a bit weird.

spec/integration/database_spec.rb
mezis: It didn't work, or rather risked leaking stuff between specs :)

The method was on `app_helper` which deals with the reusable scaffold tarball. It required keeping the clean scaffold uncompressed an ran the `rake` task there, instead of the 1-off uncompressed version used in each spec.

spec/support/build_test_app.rb
mezis: Facepalm. Yes.

README.md
mezis: Sorted.

spec/support/build_test_app.rb
danielcooper: I think this needs fixing before going out

nronas:
LGTM - Agreed with Greg the workaround is kinda itchy but not a blocker :)

PR #14, opened 2017-03-27
Don’t assume ActionDispatch::Cookies is in the middleware

danielcooper:
When generating an app for rails in api mode ActionDispatch::Cookies is not included in the stack


knaveofdiamonds:
I am sad this did not get merged :(

PR #15, opened 2017-03-31
LT-428: Don't use Rack::SslEnforcer in test environment

gregbeech:
Capybara does not play nicely with Rack::SslEnforcer and the
workarounds to actually get it to do so are horrific. As test should
only be used for actual tests and not staging environments, disabling
this middleware for the test environment only seems like a pretty
reasonable step.

### DIFF
--- a/lib/roo_on_rails/railties/http.rb
+++ b/lib/roo_on_rails/railties/http.rb
@@ -28,10 +28,13 @@ module RooOnRails
           app.config.middleware.use ::Rack::Deflater
         end
 
-        app.config.middleware.insert_before(
-          ActionDispatch::Cookies,
-          ::Rack::SslEnforcer
-        )
+        # Don't use SslEnforcer in test environment as it breaks Capybara
+        unless Rails.env.test?
+          app.config.middleware.insert_before(
+            ActionDispatch::Cookies,
+            ::Rack::SslEnforcer
+          )
+        end
       end
     end
   end
--- a/spec/integration/http_spec.rb
+++ b/spec/integration/http_spec.rb
@@ -29,8 +29,19 @@ describe 'Http rack setup' do
       end
     end
 
-    it 'inserts rack enforcer into the middleware stack' do
-      expect(middleware).to include 'Rack::SslEnforcer'
+    context 'if RAILS_ENV is not set to "test"' do
+      it 'inserts rack enforcer into the middleware stack' do
+        expect(middleware).to include 'Rack::SslEnforcer'
+      end
+    end
+
+    context 'if RAILS_ENV is set to "test"' do
+      before { ENV['RAILS_ENV'] = 'test' }
+      after { ENV['RAILS_ENV'] = nil }
+
+      it 'does not insert rack enforcer into the middleware stack' do
+        expect(middleware).to_not include 'Rack::SslEnforcer'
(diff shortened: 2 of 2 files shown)


spec/integration/http_spec.rb
```diff
@@ -29,8 +29,19 @@
...
-      expect(middleware).to include 'Rack::SslEnforcer'
+    context 'if RAILS_ENV is not set to "test"' do
+      it 'inserts rack enforcer into the middleware stack' do
+        expect(middleware).to include 'Rack::SslEnforcer'
+      end
+    end
+
+    context 'if RAILS_ENV is set to "test"' do
+      before { ENV['RAILS_ENV'] = 'test' }
+      after { ENV['RAILS_ENV'] = nil }
```
danielcooper: I'm surprised this works. Aren't we calling a rack script here by backticking out?

gregbeech: Um, yeah, I'm surprised it works too. I don't really know how or why it does... but it does. Maybe @mezis has some insight here...?

lib/roo_on_rails/railties/http.rb
```diff
@@ -28,10 +28,13 @@ class HTTP < Rails::Railtie
...
         end
 
-        app.config.middleware.insert_before(
-          ActionDispatch::Cookies,
-          ::Rack::SslEnforcer
-        )
+        # Don't use SslEnforcer in test environment as it breaks Capybara
+        unless Rails.env.test?
+          app.config.middleware.insert_before(
+            ActionDispatch::Cookies,
```
knaveofdiamonds: Can we avoid referring to `ActionDispatch::Cookies`? `--api` mode doesn't have cookies. In fact, is there some way of removing the middleware in the test env rather than adding it in in non-test envs?

spec/integration/http_spec.rb
mezis: We're "backticking out" but something we load looks up the `RAILS_ENV` we pass. Some port of Rails loading presumably :)

lib/roo_on_rails/railties/http.rb
gregbeech: We can try, but the relative ordering of this to cookies is specifically important as it interacts with it differently depending on ordering. 

If you're running in API mode how are you going to secure the Sidekiq dashboard? Or will you not have one?

knaveofdiamonds: How do you manage it in logistics-dashboard - I assume IP based? i.e. I can go to https://logistics-dashboard-production.herokuapp.com/sidekiq without any auth step, but I need to use google auth to go to https://logistics-dashboard-production.herokuapp.com/. My understanding is that the sidekiq dashboard is completely separate from whatever you do in rails.

gregbeech: We haven't secured it yet -- it's on the list. But it will use the same cookies as the main site. Hence my question really, I can't see a good way to do it without cookies.

I'll have a quick look at supporting API mode here but as this isn't introducing cookies merely fixing a bug with test mode I don't think it needs to be part of this PR.

gregbeech: Actually this concern is already addressed by #14 so I'm going to merge.

PR #16, opened 2017-04-06
Allow database migrations to have a longer timeout

gregbeech:
The current 200ms statement timeout also applies itself to migrations,
which isn’t fun when you’re trying to do things like adding indexes to
existing migrations. This PR changes the statement timeout for
migrations to a much more appropriate value.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,6 +7,10 @@ Style/FrozenStringLiteralComment:
 Style/ModuleFunction:
   EnforcedStyle: extend_self
 
+Style/PercentLiteralDelimiters:
+  PreferredDelimiters:
+    '%i': ()
+
 # Disable Metric checks for specs
 Metrics/BlockLength:
   Exclude:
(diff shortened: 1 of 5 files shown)


.rubocop.yml
```diff
@@ -7,6 +7,10 @@ Style/FrozenStringLiteralComment:
 Style/ModuleFunction:
   EnforcedStyle: extend_self
 
+Style/PercentLiteralDelimiters:
+  PreferredDelimiters:
+    '%i': ()
```
mezis: It generates arrays, so IMHO the default (prefer square brackets) is superior.

Also, we need an official Rubocop file :)

mezis:
Could we possibly get a test though?

.rubocop.yml
gregbeech: The default _was_ `()` (this issue has only just started firing) and it still suggests that `%w()` is correct. Annoyingly the behaviour actually seems to be _different_ locally vs on CodeClimate and this way forces them to be the same. I'll change to force it to `[]`.

gregbeech:
@mezis have added a test and reduced the timeout to 10 seconds. The test is pretty nasty as this task only operates in the context of `db:migrate` and friends so the only way I could really test it was to run a migration that would take longer than the default 200ms and ensure it succeeded.

PR #18, opened 2017-04-11
Fully qualify Rack constants

gregbeech:
This error handler masks application errors because it causes Ruby to
blow up with “undefined constant RooOnRails::Rack::Timeout” when an
actual error occurs and it’s running through the exception handlers.


tompave-roo:
@knaveofdiamonds Ruby is getting confused because you're inside a `***::Rack` module, and your constants are `Rack::***`. It tries to resolve them in the first `Rack` it encounters walking up the chain.

mezis:
Nicely caught, @gregbeech

gregbeech:
I'd screwed up a controller, and was wondering "why the hell am I getting this random error".

We've got a whole bunch of fixes in the gem, so might be worth releasing once this is merged.

PR #19, opened 2017-04-18
Publish lifecycle events to routemaster

gregbeech:
Here's a proposal for how publishing events could look from model classes. It's based on the code in `orderweb` but I've cleaned up the naming/implementation and made it more amenable to publishing multiple events from a single model by making multiple publishers the default.


lib/roo_on_rails/routemaster/lifecycle_events.rb
```diff
@@ -0,0 +1,49 @@
+require 'active_support/concern'
+require 'new_relic/agent'
+
+module RooOnRails
+  module Routemaster
+    module LifecycleEvents
+      extend ActiveSupport::Concern
+
+      ROUTEMASTER_EVENT_TYPE_MAP = {
```
mezis: I'd add a little comment — this maps _ActiveRecord_ events to Routemaster events?

lib/roo_on_rails/routemaster/lifecycle_events.rb
```diff
@@ -0,0 +1,49 @@
+require 'active_support/concern'
+require 'new_relic/agent'
```
mezis: Missing a require for `roo_on_rails/routemaster/publishers` which is referenced.

lib/roo_on_rails/routemaster/publisher.rb
```diff
@@ -0,0 +1,49 @@
...
+      end
+
+      def publish?
+        true
+      end
+
+      def publish!
+        return unless publish?
+
+        event_data = data # cache in case it isn't memoised
```
mezis: erm, "cache" ? "memoise"? `data` is a stub/virtual method so this doesn't actually do anything. I think you could replace lines 18-21 with just:

```ruby
routemaster.send(event, topic, url, data: recustive_stringify(event_data))
```

mezis: (with a caveat for the `nil` case of course)

lib/roo_on_rails/routemaster/publisher.rb
```diff
@@ -0,0 +1,49 @@
+module RooOnRails
+  module Routemaster
+    class Publisher
+      attr_reader :model, :event
+
+      def initialize(model, event)
```
mezis: I'd suggest dependency-injecting a `client: ::Routemaster::Client` here — otherwise people are going to stub like hell when testing publisherS?

lib/roo_on_rails/routemaster/publisher.rb
```diff
@@ -0,0 +1,49 @@
+module RooOnRails
```
mezis: Missing a require for `routemaster/client`.

lib/roo_on_rails/routemaster/publisher.rb
gregbeech: To be fair this pattern is copied from your code ;-)

But yes I did mean memoise. It could be cleaned by handling nil in the stringify method though.

lib/roo_on_rails/routemaster/publisher.rb
gregbeech: Fair

danielcooper:
Seems like a good plan.

gregbeech:
@mezis @danielcooper this is going to be "interesting" to test properly as RooOnRails itself isn't a Rails app so doesn't have any database etc. to actually test an ActiveRecord model against.

The only thing I really need that for is the `after_commit` hook, so any objections to me faking up an `after_commit` hook with the same signature on the model and checking that it calls the right thing rather than going through the whole mess of actually have a real ActiveRecord model to test against?

danielcooper:
Can you provide a nice api for doing it, and perhaps a test helper - rather than mocking it out?

danielcooper:
(not that you can test after commit when you've transactional tests anyway)

README.md
```diff
@@ -199,6 +199,16 @@ application.
...
+### Routemaster client
+
+[routemaster-client](https://github.com/deliveroo/routemaster-client) comes as a dependency with a base events publisher.
+
+To publish events onto the bus from your app, you must:
+
+* Make sure you have the `ROUTEMASTER_ENABLED` env var set to `true`.
+* Add the `publish_lifecycle_events` class method to the relevant model(s).
+* Create publisher file(s) with `url` method and others if applicable.
+* Register your publisher(s) in `config/initializers/routemaster.rb`.
```
mezis: This is a little confusing - Routemaster doesn't require publishers to register?

README.md
```diff
@@ -199,6 +199,16 @@ application.
...
 
+### Routemaster client
+
+[routemaster-client](https://github.com/deliveroo/routemaster-client) comes as a dependency with a base events publisher.
+
+To publish events onto the bus from your app, you must:
+
+* Make sure you have the `ROUTEMASTER_ENABLED` env var set to `true`.
+* Add the `publish_lifecycle_events` class method to the relevant model(s).
+* Create publisher file(s) with `url` method and others if applicable.
```
mezis: Maybe a detailed example would be useful, as we did in `README.google_oauth2.md`?

README.md
mezis: If what you mean if "configure the Routemaster client" (e.g. tell it what the bus URL is), I think we'd prefer (within Roo on Rails) if this was done automatically based on environment variables.

kplattret: No I do mean register the publisher, using `RooOnRails::Routemaster::LifecycleEvents.register` ideally in an initialiser.

I agree that the `Routemaster` config should live in `RooOnRails` and use the environment variables. I'll make that change.

lib/roo_on_rails/routemaster/lifecycle_events.rb
```diff
@@ -0,0 +1,50 @@
...
+
+      def routemaster_event_type(event)
+        ACTIVE_RECORD_TO_ROUTEMASTER_EVENT_MAP[event].tap do |type|
+          raise "invalid lifecycle event '#{event}'" unless type
+        end
+      end
+
+      module ClassMethods
+        def publish_lifecycle_events(*events)
+          events ||= %i(create update destroy)
```
danielcooper: This needs to be something along the lines of

```
    events = events.any? ? events : %i(create update destroy)
```
`*events`  will produce [] with no arguments

danielcooper: Additionally, `described_class._commit_callbacks` will give you a list of registed callbacks. You should use this in your test.

kplattret: Oh right, that's super helpful. I'll try that, thank you 🙂

spec/integration/routemaster_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+
+      context 'and ROOBUS_URL/ROOBUS_UUID are set' do
+        before do
+          ENV['ROOBUS_URL'] = 'https://routemaster.dev'
+          ENV['ROOBUS_UUID'] = 'demo'
+        end
+
+        it 'inserts Routemaster Client into the middleware stack' do
+          app.wait_start
+          expect(output).to include 'Routemaster::Client'
```
danielcooper: Why does the client have to be in the middleware stack?

lib/roo_on_rails/routemaster/publisher.rb
```diff
@@ -0,0 +1,57 @@
...
+    class Publisher
+      attr_reader :model, :event
+
+      def initialize(model, event, client: ::Routemaster::Client)
+        @model = model
+        @event = event
+        @client = client
+      end
+
+      def publish?
```
rfonglip: Why does this method always return true?

danielcooper: The intention is that it's ready to be overloaded in the clients own publishers

kplattret: The idea is that it's `true` by default and you can then overwrite this in your implementation. You can see this in the example I provide.

danielcooper:
@rfonglip 

You can either overload the publisher's `publish!` method to switch the event type if the object has it's flag set or try and tie in with paranoia's callbacks - calling publish_lifecycle_event_on_delete (although I assume you'll need to find a way for it not to trigger an update callback).

I suppose it depends on us having a standard gem we use for soft deletes?

README.md
```diff
@@ -199,6 +199,14 @@ application.
...
 A simple but secure example is detailed in `README.google_oauth2.md`.
 
+### Routemaster Client
+
+When `ROUTEMASTER_ENABLED` is set to `true` we attempt to configure [`routemaster-client`](https://github.com/deliveroo/routemaster-client) on your application. In order for this to happen you need to set the following environment variables:
+
+* `ROOBUS_URL` – the full URL of your Routemaster application (mandatory)
+* `ROOBUS_UUID` – the UUID of your application, e.g. `logistics-dashboard` (mandatory)
+
+If you then want to enable the publishing of events onto the event but, you need to set `ROUTEMASTER_PUBLISHING_ENABLED` to `true` and implement publishers as needed. An example of how to do this is detailed in [`README.routemaster_client.md`](README.routemaster_client.md).
```
scott-ad-riley: minor typo: `but` -> `bus`

kplattret: Good spot, thank you!

README.routemaster_client.md
```diff
@@ -0,0 +1,107 @@
...
+
+```ruby
+# app/publishers/base_publisher.rb
+require 'roo_on_rails/routemaster/publisher'
+
+class BasePublisher < RooOnRails::Routemaster::Publisher
+  include Rails.application.routes.url_helpers
+
+  def publish?
+    noop? || model.new_record? || model.previous_changes.any?
```
gregbeech: Could always move this into the base implementation.

README.routemaster_client.md
```diff
@@ -0,0 +1,107 @@
...
+end
+```
+
+Then create a publisher for each model with lifecycle events enabled:
+
+```ruby
+# app/publishers/order_publisher.rb
+class OrderPublisher < BasePublisher
+  def url
+    api_order_url(model, host: ENV.fetch('API_HOST'), protocol: 'https')
```
gregbeech: Could move this into the base implementation too.

lib/roo_on_rails/railties/routemaster.rb
```diff
@@ -0,0 +1,36 @@
...
+      end
+
+      private
+
+      def roobus_credentials_blank?
+        roobus_url.blank? && roobus_uuid.blank?
+      end
+
+      def roobus_url
+        ENV.fetch('ROOBUS_URL')
```
gregbeech: It strikes me that using ROUTEMASTER_URL and ROUTEMASTER_UUID would probably be cleaner names, but if people are dead set on ROOBUS_ then I guess that's fine.

README.routemaster_client.md
gregbeech: Ah, no, this one can't really move into the base implementation because sometimes there's an `api_` prefix on the routes, sometimes not, and sometimes it's different. That's why I just raised I think.

kplattret: Yeah I think this one is best to handle in the individual service apps indeed.

README.routemaster_client.md
kplattret: That sounds reasonable to me 👍

lib/roo_on_rails/railties/routemaster.rb
kplattret: I just assumed that these were prefixed with `ROOBUS_` for a good reason and didn't even question it, but it would definitely make sense to be more consistent and use `ROUTEMASTER_`. Happy to make this change if people agree.

kplattret:
It looks like the failing tests we're seeing on Travis as of this afternoon are due to an issue with the `farraday_middleware` gem (see: lostisland/faraday_middleware#165).

README.md
```diff
@@ -16,16 +16,18 @@
...
-- [Usage](#usage)
-- [Library features](#library-features)
-    - [New Relic configuration](#new-relic-configuration)
-    - [Rack middleware](#rack-middleware)
-    - [Database configuration](#database-configuration)
-    - [Sidekiq](#sidekiq)
-    - [HireFire Workers](#hirefire-workers)
-    - [Logging](#logging)
-    - [Google Oauth](#google-oauth)
+- [Configuration and usage](#configuration-and-usage)
```
mezis: Can we keep the "library usage" v "command features" sections? Users have been confused by the dual nature of `roo_on_rails` (it's both a library/framework complement _and_ a CLI, usable separately).

lib/roo_on_rails/railties/routemaster.rb
```diff
@@ -0,0 +1,36 @@
...
+        ::Routemaster::Client.configure do |config|
+          config.url = routemaster_url
+          config.uuid = routemaster_uuid
+        end
+      end
+
+      private
+
+      def bus_credentials_blank?
+        routemaster_url.blank? && routemaster_uuid.blank?
```
Crunch09: Shouldn't this be `routemaster_url.blank? || routemaster_uuid.blank?` ? (Assuming both are mandatory)

README.md
kplattret: Absolutely! I was a bit confused by this table of content myself, partly because it changed while I was working on this PR. I'll rename **"Configuration and usage"** to **"Library usage"** 👍

lib/roo_on_rails/railties/routemaster.rb
kplattret: Yes indeed, good spot, thank you.

PR #20, opened 2017-04-20
Adds a custom logging formatter in the vein of TaggedLogging

gregbeech:
While looking at getting things to play nicely with Scalyr and its k=v
field extraction, I wanted to move the burden of formatting the logs
correctly out of (many places in) the app and into a custom formatter.

The `ActiveSupport::TaggedLogging` looked pretty much like what I
wanted to achieve, but instead of having nameless `[tag]` prefixes on
the message have `k=v` pairs. As there could be numerous attributes it
seemed more readable to have them as a suffix than a prefix so they
didn’t push the message itself off too far.

I've looked at putting the `msg` argument to the logger in a `msg=`
key/value pair, but the trouble is some libraries already format their
logs in logfmt style and this sweeps up those pairs into a string which
is bad. I _could_ do something hacky like parse the string and then scan
for keys which have a `true` value (meaning, typically, they were just
a lone word) and concat those things as the message... but... that 
seems ropey so I've just left it alone.

_Note 1: I haven't written tests yet; wanted to get the concept approved
before I did that._

_Note 2: I have looked at [Semantic Logger](https://github.com/rocketjob/semantic_logger)
but I'm not sure if it's overkill. If we want to use it then it would be easy
enough to repurpose this as an appender for that in logfmt format. Let
me know your thoughts!_


lib/roo_on_rails/context_logging.rb
```diff
@@ -0,0 +1,80 @@
...
+      end
+
+      private
+
+      def context_text
+        context = current_context
+        return nil unless context.any?
+
+        merged_context = context.each_with_object({}) { |ctx, acc| acc.merge!(ctx) }
+        ' -- ' + merged_context.map { |k, v| "#{k}=#{v}" }.join(' ')
```
mezis: It looks like you're trying to implement [logfmt](https://blog.codeship.com/logfmt-a-log-format-thats-easy-to-read-and-write/), also known as "the Heroku logging style". I'm in favour, but maybe there's a library out there?

If you want to do it yourself — you're missing escaping. A classic seems to be JSON-escaping the value.

lib/roo_on_rails/context_logging.rb
```diff
@@ -0,0 +1,80 @@
...
+      end
+
+      def context(**context)
+        push_context(**context)
+        yield self
+      ensure
+        pop_context
+      end
+
+      def push_context(**context)
```
mezis: Explicit push/pop as public methods sounds ... dangerous. What's wrong with the `context` method above?

lib/roo_on_rails/context_logging.rb
```diff
@@ -0,0 +1,80 @@
...
+  #
+  # You can also use `push_context` and `pop_context` in before/after blocks.
+  #
+  #   class ApplicationController < ActionController::Base
+  #     before_action { logger.push_context(user_id: current_user.id) }
+  #     after_action { logger.pop_context }
+  #   end
+  module ContextLogging
+    module Formatter
+      def call(severity, timestamp, progname, msg)
```
mezis: Other libraries (e.g. [Slog](https://github.com/sczizzo/slog)) or platforms (e.g. [this](https://github.com/sirupsen/logrus#example)) tend to follow this format instead:

```
$logger.with(key: 'value').info("human readable message")
```

lib/roo_on_rails/context_logging.rb
```diff
@@ -0,0 +1,80 @@
...
+        super(severity, timestamp, progname, "#{msg}#{context_text}")
+      end
+
+      def context(**context)
```
mezis: Maybe call this `with`?

lib/roo_on_rails/context_logging.rb
```diff
@@ -0,0 +1,80 @@
...
+        ' -- ' + merged_context.map { |k, v| "#{k}=#{v}" }.join(' ')
+      end
+    end
+
+    def self.new(logger)
+      # Ensure we set a default formatter so we aren't extending nil!
+      logger.formatter ||= ActiveSupport::Logger::SimpleFormatter.new
+      logger.formatter.extend(Formatter)
+      logger.extend(self)
+    end
```
mezis: So much `extend`. Any reason not to use delegation?

lib/roo_on_rails/context_logging.rb
gregbeech: Nothing really... I was just copying what `ActiveSupport::TaggedLogging` has. It makes some things like before/after action blocks slightly cleaner as you don't have to worry about error handling in an around_action block. But happy to remove as it clearly _is_ running with scissors.

lib/roo_on_rails/context_logging.rb
gregbeech: Mainly because this is the way `ActiveSupport::TaggedLogging` is implemented, and so I figured I'd follow the same approach.

lib/roo_on_rails/context_logging.rb
gregbeech: Yeah, that sounds nicer.

lib/roo_on_rails/context_logging.rb
gregbeech: Can't easily support the dot syntax as there's no 'end' to the block, so it'd be something like:

```ruby
logger.with(key: 'value') { logger.info('human readable message') }
```

If it's something you _really_ want then I could do it though.

lib/roo_on_rails/context_logging.rb
gregbeech: Interesting, I didn't realise that format actually had a specific definition. Will take a look.

gregbeech: Sadly although there's a `logfmt` gem it only _parses_ the format but can't generate it. There also doesn't seem to be any formal spec ([this Go package is apparently about as close as it gets](https://godoc.org/github.com/kr/logfmt)). It doesn't look very complex so I'll write a bit of custom formatting code.

mezis: Well, there's this:
https://github.com/MitinPavel/logfmt_marshalling

But it doesn't looks exactly maintained :D

gregbeech: Yep, and I'm also not convinced it's _correct_ either. Though sadly there is no spec so it's very hard to tell... :-/

lib/roo_on_rails/context_logging.rb
gregbeech: This was actually pretty easy to do so I did it.

lib/roo_on_rails/logfmt.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      private
+
+      def dump_value(v)
+        str = case v
+              when String then v
+              when Symbol then v.to_s
+              else v.respond_to?(:to_json) ? v.to_json : v.inspect
+              end
+        escape(str)
```
mezis: If it's been dumped to JSON, it's already escaped (by JSON)?
Incidentally that gives you a neat way to do escaping for strings: just dump them to JSON.

ie. this should be enough, I think:

```ruby
dev dump_value(v)
  v.respond_to?(:to_json) ? v.to_json : v.inspect
end
```

gregbeech: Dumping to JSON gives you unescaped JSON, e.g. a hash of `{ h: { foo: 'bar' } }` would come out as `h={"foo":"bar"}` in the output. As this contains quotes it needs to be a qvalue and be written to the log as `h="{\"foo\":\"bar\"}"` (where the `\"` here is two separate characters, not a single quote character being displayed as escaped).

Also, dumping to JSON would quote string like `hello` to `"hello"` whereas it's cleaner to have them shown as `hello`, and you have to be careful with numbers as if they've been stored in TLS then the integer `2` might actually be a string, so you'd end up outputting it as `"2"` if you didn't check whether it was an ident.

gregbeech: It's _really_ hard to write specs for this stuff and be sure you've got it right because you've got the conflict of Ruby's embedded escaping with the fact you actually want literally escaped characters in the output. I've been doing a whole bunch of testing by using it in an app and making sure the output is correct in the real log output too ;-)

gregbeech:
Groan... and of course it doesn't work with Rails 3 out of the box...

lib/roo_on_rails/context_logging.rb
```diff
@@ -0,0 +1,97 @@
...
+      def clear_context!
+        current_context.clear
+      end
+
+      private
+
+      def current_context
+        # We use our object ID here to avoid conflicting with other instances
+        thread_key = @thread_key ||= "roo_on_rails:logging_context:#{object_id}".freeze
+        Thread.current[thread_key] ||= []
```
danielcooper: I'm a bit worried about this. I don't know if the object_id sorts it out.
https://github.com/steveklabnik/request_store

I also think this might be surprising when using promises. 

Would you be able to get the same effect (in rails) by ditching the thread and setting and clearing the log context via a middleware (at the top and the bottom) of the request chain?

gregbeech: TBH I did it this way because it's the way it's implemented in [`ActiveSupport::TaggedLogging`](https://github.com/rails/rails/blob/f1c5b0bd0a6211286bfbb8f46991de325d45a2e5/activesupport/lib/active_support/tagged_logging.rb#L47) and figured that matching the behaviour of the built in tagging thing (which is effectively what this is doing) would follow the principle of least astonishment (i.e. your context will do the exact same thing as your tags).

Once you start doing cross-thread things with context it gets _very_ difficult. At my last company we spent days/weeks trying to get cross-thread context to flow in Scala because everything we did was async and it never quite worked correctly. It's just too easy to 'leak' stuff. It's also not necessarily obvious when using promises that the context _should_ automatically flow, as promises are arguably a different context.

mezis: @danielcooper: the problem with the "middleware" approach is you still need to deal with e.g. Sidekiq. It also effectively precludes using async/threading in controllers.

A "typical" solution if we want to support logging continuity with e.g. promises is to perform context-passing — this is what [golang](https://blog.golang.org/context) favours for instance. Feels overkill for now.

We _do_ need to address your point though: as is, the code will break with servers that reuse threads (e.g. Thin, as per `request_store`'s example), because the context might not be reset properly on some exceptions. The `ensure` block need to do [this dance](https://github.com/mperham/connection_pool/blob/master/lib/connection_pool.rb#L59-L70), I believe.

gregbeech: > as is, the code will break with servers that reuse threads

So you're saying that Rails 5's `ActiveSupport::TaggedLogging` (which has the exact same implementation) is also broken? That would seem surprising... Am I missing something?

tompave-roo: On the middleware approach, @samlevy had to debug a pretty nasty leak issue with request store, a while ago. Even more worrying is that it seemed to happen with Unicorn, IIRC.

gregbeech: @mezis @danielcooper I've now made the `with` implementation interrupt-safe in the same way as connection pool (which seems to be the documented approach anyway).

gregbeech:
@mezis I've added CHANGELOG and README entries now.

I did spend quite a while trying to get a railtie to automatically set the logger up, but while it seems it's easy to swap the logger out in `application.rb` by the time the railties run it seems to be too late and if you change the logger it just gets ignored. I tried a bunch of the different hooks but none of them seemed effective. I'm sure it's possible... but I started to feel like I was spending more time on it than it was actually worth. Would be nice to add as a separate PR if anybody knows the magic incantations to swap the logger in a railtie though.

danielcooper:
@gregbeech 

https://github.com/shadabahmed/logstasher/blob/master/lib/logstasher/railtie.rb

I reckon `:before => :load_config_initializers` will do it

gregbeech:
@danielcooper it works if I set `Rails.logger` but if I set `app.config.logger` which the documentation suggests you're supposed to then it just gets ignored. I think it's worth merging this now (especially as it fixes the Sidekiq issue) and then can add a Railtie as a separate thing later.

PR #22, opened 2017-04-27
HOTFIX: Use correct SLA class name

gregbeech:
This could really do with some tests, but I’m not quite sure how to
test it and this is causing errors in production so I’d prefer to patch
it quickly and then work out how to test it later.

@danielcooper maybe you have some thoughts on how to test?


danielcooper:
Will have to boot the app and hit the endpoint i suppose. 

LGTM, whoops

PR #25, opened 2017-05-03
[CX-XXX] add rails 5.1 support

Crunch09:


spec/support/build_test_app.rb
```diff
@@ -39,8 +39,13 @@ def ensure_scaffold
...
         end
 
+        if Rails::VERSION::MAJOR < 5
+          append_to_file scaffold_dir.join('Gemfile'), %{
+            gem 'puma', '~> 3.0'
+          }
+        end
+
         append_to_file scaffold_dir.join('Gemfile'), %{
-          gem 'puma', '~> 3.0'
```
Crunch09: Removed this for Rails < 5 because Rails 5.0.x is [already including puma ~> 3.0](https://github.com/rails/rails/blob/5-0-stable/railties/lib/rails/generators/app_base.rb#L185) and Rails 5.1.x is [including puma ~> 3.7](https://github.com/rails/rails/blob/5-1-stable/railties/lib/rails/generators/app_base.rb#L194) which causes a conflict with this line.

sideshowcoder:
LGTM, gonna merge that.

PR #26, opened 2017-05-05
Allow services which do not necessarily have ActiveRecord

will-clarke:
I'm trying to use `roo_on_rails` without a db but it blows up as we're assuming everyone's going to be using `ActiveRecord`


lib/roo_on_rails/railties/database.rb
```diff
@@ -1,7 +1,7 @@
 module RooOnRails
   module Railties
     class Database < Rails::Railtie
-      if ActiveRecord::VERSION::MAJOR >= 4
+      if defined?(ActiveRecord) && ActiveRecord::VERSION::MAJOR >= 4
```
sideshowcoder: Maybe I'm wrong but the railsy way to do this would be:

```ruby 
ActiveSupport.on_load :active_record do 
   # all the active record things
end
```

lib/roo_on_rails/tasks/db.rake
```diff
@@ -23,5 +23,6 @@ end
   db:migrate:down
   db:rollback
 ].each do |task|
+  break unless defined?(ActiveRecord)
```
sideshowcoder: I feel like we can ignore this entire file unless we have ActiveRecord right? So wrapping it in a `if defined?(ActiveRecord)` would almost feel better? Or at least those task enhancing functionality would be best wrapped in this `if` block.

lib/roo_on_rails/railties/database.rb
will-clarke: Oh, okay, I didn't know about `ActiveSupport.on_load :active_record` 👍

gregbeech: How does `ActiveSupport.on_load` work with Railties though...? I'm not sure I even want to hazard a guess!

gregbeech:
👍  looks good

pedrocunha:
Would be cool if we could write some spec for this?

lib/roo_on_rails/railties/database.rb
sideshowcoder: Something like that I guess... https://github.com/sideshowcoder/activerecord_translatable/blob/master/lib/activerecord_translatable.rb#L10-L18 (oh boy that was a long time ago please don't look at the other code)

mezis: `on_load` is well documented in Rails 5+ and seems to exist before. I guess it's worth a try.

mezis:
@will-clarke @gregbeech @pedrocunha @sideshowcoder — I took the liberty to add a spec for this, switch to `on_load`, and add a CHANGELOG entry. Mind having another look?

PR #28, opened 2017-05-18
Added `enforce_admins` to the github protections call

martyphee:
Updated because of api changes.  `enforce_admins` is a required parameter.  

Check is still failing though.  Not ready to merge.


gregbeech:
@martyphee I made a few changes to get the 'fix' part of it working, so now it fixes the GitHub branch protection settings properly as well as passing the checks:

image

gregbeech:
The branch protection settings are rather draconian after this has run. Too draconian? Maybe. I'm still undecided about whether to dismiss approvals on commit because it seems a fairly frequent thing to approve PRs that need a few minor changes and trust that the only changes made will be those minor ones, so we might want to relax this restriction in future.

image

lib/roo_on_rails/checks/github/branch_protection.rb
```diff
@@ -35,7 +36,9 @@ def fix
 
         def ensure_status_checks!
           status_checks = protection[:required_status_checks] || {}
-          fail! 'status checks do not include admins' unless status_checks[:include_admins]
+          enforce_admins = protection[:enforce_admins]
+          fail! 'status checks are not enforced for admins' unless enforce_admins &&
+                                                                   enforce_admins[:enabled]
```
nronas: i'm probably missing something here, but enforce_admins is a boolean, so it will not respond to `#[]`

gregbeech: Weirdly in spite of the name it's actually a hash!

```
"enforce_admins": {
    "url": "https://api.github.com/repos/octocat/Hello-World/branches/master/protection/enforce_admins",
    "enabled": true
  },
```

([source](https://developer.github.com/v3/repos/branches/))

nronas: 👍

PR #29, opened 2017-05-18
Enable more heroku app names

shaley91:
This PR allows more permutations of heroku app names.

This helps when a github repo name is quite long as heroku limits app names to be 30 characters.


mezis:
I'm not sure I agree with this one. The point of this check is to allow us to programmatically match a repo to apps — by allowing "parts" of the repo name to be used, that becomes impossible to do reliably.

For instance: `loris` v. `loris-hook`.

I'd be in favour, when this check fails, to instead rename either the app or the repo.
(e.g. a repo doesn't need to be called `fubar-service` when `fubar` is enough?)

PR #33, opened 2017-06-01
SETI-74: Datadog app metrics

mezis:
As it says on the tin.

===

Jira story [#SETI-74](https://deliveroo.atlassian.net/browse/SETI-74) in project *Software Engineering Tools and Infrastructure*:

> `roo_on_rails` should detect whether apps 
> 
> - have metrics logging enabled, and
> - are hooked up to the logplex/datadog bridge
> 
> (and fix things if not)


lib/roo_on_rails/checks/environment.rb
```diff
@@ -4,13 +4,15 @@
...
 require 'roo_on_rails/checks/sidekiq/settings'
+require 'roo_on_rails/checks/heroku/drains_metrics'
 
 module RooOnRails
   module Checks
     class Environment < EnvSpecific
       requires GitHub::BranchProtection
       requires Heroku::PrebootEnabled
       requires Sidekiq::Settings
+      requires Heroku::DrainsMetrics
```
gregbeech: Being picky, would be nice to have these in alphabetical order (and maybe the `require`s too).

lib/roo_on_rails/checks/heroku/metrics_bridge_configured.rb
```diff
@@ -0,0 +1,73 @@
...
+            token_var    => SecureRandom.hex(16),
+            app_list_var => app_list.to_a.join(','))
+        end
+
+        def app_list_var
+          'ALLOWED_APPS'
+        end
+
+        def tags_var
+          "%s_TAGS" % app_name.upcase
```
gregbeech: Should we also `.tr('-', '_')` to make it more "environment variably"? `FOO-BAR_TAGS` seems odd compared to `FOO_BAR_TAGS`.

mezis: Agree - but that's not how the `roo-dd-bridge` app works. `FOO-BAR_TAG` is unfortunately the format it expects!

lib/roo_on_rails/checks/environment.rb
mezis: <scratches head>

Why would you want them in alphabetical order?
(happy to though)

Gemfile.lock
```diff
@@ -0,0 +1,251 @@
+PATH
```
tompave-roo: This wasn't in version control until now. Has anything changed?

gregbeech: AFAIK this shouldn't really be in here for libraries as they should install the latest thing that's permitted by the `Gemfile` (as that's what'll happen in apps who use it). It's apps that should have the lockfile.

mezis: I added it lazily because Code Climate was unhappy without. Back to the drawing board.

mezis: (the appraisal lockfiles probably shouldn't be checked in either ... i'll check how easy they'd be to remove)

tompave-roo: I think it can be part of the git repo, as long as it's excluded from the gem bundle that gets installed.

.rubocop.yml
```diff
@@ -28,6 +28,16 @@ Metrics/LineLength:
...
 Style/Documentation:
   Enabled: false
+
+
+Style/FormatString:
+  Enabled: false
+Style/DotPosition:
+  Enabled: false
+Style/TrailingWhitespace:
+  Enabled: false
```
gregbeech: Should we really be disabling these things? The first two are trivial to write in a 'compliant' way and removes the discussion about where dots should go, and the last one means that you won't get a load of "not really changed" lines the next time somebody with a "trim trailing whitespace on save" setting in their editor (which most of us use) saves a file.

mezis: This is exactly why I'd love us to have a `.rubocop.yml` that everyone uses ^_^
I only added those things so that Code Climate would shut up on things I consider non-issues. The former isn't mentioned in our guidelines; the latter is, so I'll fix that — in a separate PR to keep this clean though.

gregbeech: They're things _you_ consider to be non-issues... but for example _I_ certainly consider trailing whitespace to be a considerable issue. Dot position I don't hugely care about which choice we make, but I do prefer consistency to inconsistency.

Look at it this way: What if everybody disabled the couple of checks _they_ considered non-issues?

mezis: Totally fair - which is why I'm addressing the whitespace one :)

gregbeech: Fair enough. I've approved, so feel free to merge.

You know there's a setting in your editor for that whitespace thing though ;-)

PR #35, opened 2017-06-05
LT-596: Add Datadog integration and instrument Sidekiq

gregbeech:
* Adds a statsd client to roo_on_rails; it uses the `STATSD` global if it's already defined, otherwise builds a Datadog statsd client. The `STATSD_HOST` and `STATSD_PORT` env vars can be used to change where it sends metrics to.

* Inserts the Sidekiq metrics middleware if Sidekiq Pro is available. I don't hugely like this middleware's metrics as it puts the job name in the key rather than as a tag so they can't be aggregated, but it's better than nothing until we write a tagged version.

* Adds a `RooOnRails::Sidekiq::MetricsWorker` job which can be scheduled to give you decent metrics about queue sizes and latency, process counts, and auto-scaling metrics. At the moment this is _not_ automatically scheduled; that can be done in a later PR.


lib/roo_on_rails/railties/sidekiq.rb
```diff
@@ -28,6 +29,20 @@ def config_sidekiq
         end
       end
 
+      def config_sidekiq_metrics
+        begin
+          require 'sidekiq/middleware/server/statsd'
+        rescue LoadError
+          return # metrics are only available in sidekiq pro
```
danielcooper: Might want to print a warning here?

gregbeech: Yeah I was debating that, but as this is an open-source library not everybody will have Sidekiq Pro. Then again, using non-Pro for commercial stuff is a bad idea as it can lose jobs, so maybe it's a reasonable warning anyway.

gemfiles/rails_5.gemfile.lock
```diff
@@ -146,33 +149,33 @@ GEM
...
-    rails (5.0.2)
-      actioncable (= 5.0.2)
-      actionmailer (= 5.0.2)
-      actionpack (= 5.0.2)
-      actionview (= 5.0.2)
-      activejob (= 5.0.2)
-      activemodel (= 5.0.2)
-      activerecord (= 5.0.2)
-      activesupport (= 5.0.2)
+    rails (5.1.1)
```
Crunch09: Shouldn't this still be `5.0.x`? Because we already have [rails_5_1.gemfile.lock](https://github.com/deliveroo/roo_on_rails/blob/master/gemfiles/rails_5_1.gemfile.lock) for the 5.1 case.

gregbeech: Yep, good spot! The 5.0 appraisal spec wasn't restrictive enough. Will fix.

PR #36, opened 2017-06-13
Rails-api fixes

gurkanoluc:
- `ActionDispatch::Cookies` is not used in api only Rails projects, so `insert_before` that uses it fails. `Rack::Head` is common in normal rails and api only and it inserts the middleware to a place it would still continue working.
- `break` to early return from Sidekiq initializer fails with `LocalJumpError` instead of trying to return early this just wraps the functions we would like to call in `if` block.

### DIFF
--- a/lib/roo_on_rails/railties/http.rb
+++ b/lib/roo_on_rails/railties/http.rb
@@ -17,7 +17,10 @@ module RooOnRails
           ::Rack::Timeout
         )
 
-        middleware_to_insert_before = defined?(::Rack::Head) ? ::Rack::Head : ::ActionDispatch::Cookies
+        middlewares = app.config.middleware.to_a.map { |c| c.klass.to_s }
+        middlewares.each { |p| puts p.klass.to_s }
+        
+        middleware_to_insert_before = middlewares.include?('Rack::Head') ? ::Rack::Head : ::ActionDispatch::Cookies
 
         # This needs to be inserted low in the stack, before Rails returns the
         # thread-current connection to the pool.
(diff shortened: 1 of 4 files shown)


lib/roo_on_rails/railties/http.rb
```diff
@@ -17,10 +17,12 @@ class HTTP < Rails::Railtie
           ::Rack::Timeout
         )
 
+        middleware_to_insert_before = Rails::VERSION::MAJOR < 4 ? ::ActionDispatch::Cookies : ::Rack::Head
```
knaveofdiamonds: I assume this works implicitly because api mode doesn't exist before version 5, and we don't care if we insert before `Rack::Head` in all our services? I guess my question is - is it better to do it this way, or check for and switch on the existence of `ActionDispatch::Cookies` explicitly? (it could be that there is no difference)

gurkanoluc: Intention is instead of using `ActionDispatch::Cookies` using `Rack::Head` which is common in `Rails 4`, `Rails 5` and `Rails 5 Api only` but not in `Rails 3`. But we have `ActionDispatch::Cookies` for Rails 3.

spec/support/build_test_app.rb
```diff
@@ -63,7 +63,7 @@ def ensure_scaffold
         scaffold_dir.rmtree unless @keep_scaffold
         self
       rescue => e
-        binding.pry
+        # binding.pry
```
mezis: If you're removing the breakpoint (which makes sense!) you probably want to remove the blanket rescue as well?

PR #40, opened 2017-07-03
Don't Assume ActiveRecord is defined

will-clarke:
SafeTimeouts currently just clears the ActiveRecord connection cache.
We don't want this if we haven't defined ActiveRecord.


mezis:
Looks good, but we're missing a test case for this?

PR #41, opened 2017-07-06
📖 Require a playbook in the repo

jphastings:
Adds a check to the `roo_on_rails` command line app which requires a `PLAYBOOK.md` file within your repo, detailing how to deal with issues the service might have.

### DIFF
--- a/lib/roo_on_rails/checks/documentation/playbook.rb
+++ b/lib/roo_on_rails/checks/documentation/playbook.rb
@@ -1,0 +1,23 @@
+require 'roo_on_rails/checks/base'
+
+module RooOnRails
+  module Checks
+    module Documentation
+      class Playbook < Base
+        LOCATION = 'PLAYBOOK.md'.freeze
+
+        def intro
+          'Looking for an on-call Playbook...'
+        end
+
+        def call
+          if File.exist?(LOCATION)
+            pass 'playbook found, legion on-call engineers thank you.'
+          else
+            fail! "please create a playbook at #{LOCATION}."
+          end
+        end
+      end
+    end
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Pre-release
+
+Features:
+
+- Adds check for a `PLAYBOOK.md` file, which should detail how to deal with issues which might occur with the service.
+
 # v1.5.0 (2017-06-19)
 
 Features:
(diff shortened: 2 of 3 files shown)


gregbeech:
IMHO it would be better in the root as `PLAYBOOK.md` -- that's what I've done in identity as when there's an incident the last thing you want is to be digging around in folders looking for it, you want it right there and in ALL CAPS!

lib/roo_on_rails/checks/documentation/playbook.rb
```diff
@@ -0,0 +1,23 @@
...
+    module Documentation
+      class Playbook < Base
+        LOCATION = 'PLAYBOOK.md'.freeze
+
+        def intro
+          'Looking for an on-call Playbook...'
+        end
+
+        def call
+          if File.exist?(LOCATION)
```
danielcooper: Bonus points if the script offers to generate one with a nice template?

jphastings: I was thinking that, I figured the first PR to make this a requirement would be good, then a second that we can argue over for the structure :P

gregbeech: I copied @danielcooper's structure of:

```
# <service name> Playbook

<intro text>

## Useful Links

<links to Heroku, New Relic, Papertrail, etc.>

## Known Issues

### <alert title>

<how to handle the alert>
```

Seemed like quite a good start anyway.

tompave-roo:
I had a look at this this morning and I'm really not sure if it's something I would find useful.

My fear is that most people would just add an empty playbook file to make the check pass and then go on with their work. You could stat the file to check that is't not empty, but that doesn't really give you any guarantee on the quality of the content either. The key issue, I think, is that when you start work on a new project you only have a very hazy idea of the failure modes at best, and it's just not possible to write recovery rules. The content of the playbook file will necessarily have to evolve over time, which means that people can just as well forget to keep it up-to-date (it happens for readmes!). So forcing people to generate a playbook at the beginning of a project isn't really going to provide significant real-world benefits.

I had a chat with Amy and she likes the idea of tracking which applications/repos have playbooks and which don't. I can see why that would be helpful, but I would say that it's easier to identify _missing_ files than _empty_ files, and let's not even talk about identifying non-empty but low-quality files.

If we decide to keep this feature, I like @danielcooper's comment on generating a template playbook instead. Sounds more useful than just nagging people because the file is missing.

jphastings:
Hey @tompave, I completely agree — I fully plan to submit a PR with a template playbook (in fact I'm writing it right now on the top of this PR), but I reckon it'll be contentious, so I'd like to get the feature in first :)

As to the fact that people may not complete the playbook effectively; that's always a risk, people may also not read the guidelines, make sensible design decisions or turn up to work but we have faith in the skills, intellect and motivation of the engineers we work with ☺️.

This check allows us to take away that fear-of-a-blank-page that can intimidate some folks away from writing a playbook, and we can improve the template to cover more and more to help out.

PR #44, opened 2017-07-12
SETI-395: Add google auth support

tompave-roo:
Adds optional support for Google Oauth2. Configuration instructions in the diff for the Readme.

More in detail, we automatically configure and inject the `OmniAuth` middleware, and then add another custom middleware that does nothing for normal requests, but acts as terminator for the Oauth callbacks. The idea is to try to remove the need to sprinkle configuration and bits of logic here and there, for example in the Rails router.

The custom middleware requires the application to provide an authentication strategy that must return a valid Rack response. The user is responsible for providing this strategy, because each application can store the user data from the Oauth payload in different ways.

The generated Rails initializer contains examples on how to provide an authentication strategy for this custom middleware.


gregbeech:
Somewhat confused... why are we adding this when the identity service offers SSO packaged up for you? https://github.com/deliveroo/identity-service/pull/194

mezis:
> Somewhat confused... why are we adding this when the identity service offers SSO packaged up for you

@gregbeech, I think the reasoning is that for some internal/fabric services it won't be acceptable to depend on the identity service being up. The use case being, some of those apps are required to diagnose/mitigate services going down, so they should live "outside" the service federation if possible.

E.g. we'll probably use this in the "changelog" app (to replace `roo-deploys`), which must be self-contained and up at all times (_particularly_ if `identity-service` is down!)

In other words — this is mainly for apps at a "lower layer" of our arch stack diagram (fabric), which I think should not depend "up" (to identity, which straddles the "edge" and "domain" layers).

mezis:
Extra note for context — while I think we're fine to start with Google OAuth for this, I'd like to add Github OAuth. Given the use cases, it's _way_ easier to manage e.g. an "on-call" team in Github, and configure Omniauth to allow organisation=deliveroo and team=oncall, than to do something similar with Google Apps auth.

gregbeech:
OK I guess that makes sense. As you say, using GitHub probably makes more sense than Google for this (though GitHub will likely be slaved to Google soon, as we're trying out the business account, so I have no idea if that will put a spanner in the works).

Should we add some comments to this that it is not for general purpose use, and that in general you should use the identity service for authentication needs?

gregbeech:
FYI what we've started discussing in InfraSec is doing something like this:

- Namely becomes the 'source of truth' for the org chart, groups, etc.
- Active Directory (or similar) takes group data from there and federates it
- Identity Service, GitHub, etc. synchronise groups with AD/similar

Note that this was literally a 10 minute discussion, not any kind of concrete plan, so would be good to get you & maybe some others from SETI in on the planning so we can make security groups available wherever makes sense without any onboarding/offboarding woes.

PR #46, opened 2017-07-19
better Sidekiq setup through environment variables

humzashah:
- adds `default` to default to queues
- lets you configure custom Sidekiq queue names along with permitted latency values for proper monitoring and scaling
- lets you specify which environments your checks run on
- makes relevant changes to README

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,8 @@ When `SIDEKIQ_ENABLED` is set we'll:
 The following ENV are available:
 
 - `SIDEKIQ_ENABLED`
+- `SIDEKIQ_QUEUES` - comma-separated custom queue names; if not specified, default queues are processed which are defined [here](./lib/roo_on_rails/sidekiq/settings.rb)
+- `SIDEKIQ_PERMITTED_LATENCY_VALUES` - our default queues come with a permitted latency value which is used in monitoring and Hirefire scaling. For your custom queues, you will need to specify the permitted latency in this format: `QUEUE_NAME:VALUE:UNIT` e.g. `queue-one:1:day,queue-two:3:minutes,queue-three:2:seconds`
 - `SIDEKIQ_THREADS` (default: 25) - Sets sidekiq concurrency value
 - `SIDEKIQ_DATABASE_REAPING_FREQUENCY` (default: 10) - For sidekiq processes the
   amount of time in seconds rails will wait before attempting to find and
@@ -204,9 +206,18 @@ Run the following from your app's top-level directory:
 bundle exec roo_on_rails
 ```
 
-You will (currently) need to have admin privileges on the
-`roo-dd-bridge-production` application for this to work. This will be addressed
-eventually.
+That command will sequentially run a number of checks. For it to run successfully, you will need:
+
+- a GitHub API token that can read your GitHub repository's settings placed in `~/.roo_on_rails/github-token`
+- the Heroku toolbelt installed and logged in
+- admin privileges on the `roo-dd-bridge-production` (this will be addressed eventually)
+- checks are run sequentially for staging and then for production. The process halts at any non-fixable failing check. To process only specific environments, you can set a config variable while running the command, like so:
+
+```
+ROO_ON_RAILS_ENVIRONMENTS=staging,production bundle exec roo_on_rails $(: this is the default behaviour)
+ROO_ON_RAILS_ENVIRONMENTS=staging bundle exec roo_on_rails $(: run checks only on staging)
+ROO_ON_RAILS_ENVIRONMENTS=production bundle exec roo_on_rails $(: run checks only on production)
+```
 
 
 ### Description
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ When `SIDEKIQ_ENABLED` is set we'll:
 The following ENV are available:
 
 - `SIDEKIQ_ENABLED`
-- `SIDEKIQ_QUEUES` - comma-separated custom queue names ordered by increasing permitted latency; if not specified, default queues are processed which are defined [here](./lib/roo_on_rails/sidekiq/settings.rb). For accurate health reporting and scaling for your custom queue names, you can specify their permitted latency within this variable e.g. `within5seconds,queue-one:10seconds,queue-two:20minutes,queue-three:3hours,queue-four:1day,default`. For non-default queue names that don't follow the `withinXunit` pattern, you will **need** to specify the permitted latency otherwise the initialization will error.
+- `SIDEKIQ_QUEUES` - comma-separated custom queue names; if not specified, default queues are processed which are defined [here](./lib/roo_on_rails/sidekiq/settings.rb). For accurate health reporting and scaling for your custom queue names, you can specify their permitted latency within this variable e.g. `within5seconds,queue-one:10seconds,queue-two:20minutes,queue-three:3hours,queue-four:1day,default`. For non-default queue names that don't follow the `withinXunit` pattern, you will **need** to specify the permitted latency otherwise the initialization will error.
 - `SIDEKIQ_THREADS` (default: 25) - Sets sidekiq concurrency value
 - `SIDEKIQ_DATABASE_REAPING_FREQUENCY` (default: 10) - For sidekiq processes the
   amount of time in seconds rails will wait before attempting to find and
--- a/README.md
+++ b/README.md
@@ -213,9 +213,14 @@ That command will sequentially run a number of checks. For it to run successfull
 - checks are run sequentially for staging and then for production. The process halts at any non-fixable failing check. To process only specific environments, you can set a config variable while running the command, like so:
 
 ```
-ROO_ON_RAILS_ENVIRONMENTS=staging,production bundle exec roo_on_rails $(: this is the default behaviour)
-ROO_ON_RAILS_ENVIRONMENTS=staging bundle exec roo_on_rails $(: run checks only on staging)
-ROO_ON_RAILS_ENVIRONMENTS=production bundle exec roo_on_rails $(: run checks only on production)
+# the default behaviour:
+ROO_ON_RAILS_ENVIRONMENTS=staging,production bundle exec roo_on_rails
+
+# run checks only on staging:
+ROO_ON_RAILS_ENVIRONMENTS=staging bundle exec roo_on_rails
+
+# run checks only on production:
+ROO_ON_RAILS_ENVIRONMENTS=production bundle exec roo_on_rails
 ```
 
 
--- a/README.md
+++ b/README.md
@@ -112,8 +112,7 @@ When `SIDEKIQ_ENABLED` is set we'll:
 The following ENV are available:
 
 - `SIDEKIQ_ENABLED`
-- `SIDEKIQ_QUEUES` - comma-separated custom queue names; if not specified, default queues are processed which are defined [here](./lib/roo_on_rails/sidekiq/settings.rb)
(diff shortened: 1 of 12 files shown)


lib/roo_on_rails/sidekiq/queue_latency.rb
```diff
@@ -17,7 +17,17 @@ def normalised_latency
         metric.round(3)
       end
 
+      private
+
       def permitted_latency
+        entries = ENV.fetch('SIDEKIQ_PERMITTED_LATENCY_VALUES', '').split(',')
+        entry = entries.find { |e| e.start_with?(queue.name) }
```
gregbeech: Would probably be better to do this once and memoise as a `Hash` rather than doing the parsing each time.

lib/roo_on_rails/sidekiq/settings.rb
```diff
@@ -1,16 +1,19 @@
...
-          monitoring
-          realtime
-          within1minute
-          within5minutes
-          within30minutes
-          within1hour
-          within1day
-        ).freeze
+        env_key = 'SIDEKIQ_QUEUES'
+        ENV.key?(env_key) ? ENV[env_key].split(',') : DEFAULT_QUEUES
```
gregbeech: I'd probably `.map(&:trim)` after the `split` just to be on the safe side, otherwise I can see people spending ages wondering why their queue isn't being picked up because Sidekiq is looking for ` my_queue`.

lib/roo_on_rails/version.rb
```diff
@@ -1,3 +1,3 @@
 module RooOnRails
-  VERSION = '1.7.0'.freeze
+  VERSION = '1.8.0'.freeze
```
gregbeech: We should do version/release separately I think is the general approach... @mezis ?

README.md
```diff
@@ -112,6 +112,8 @@ When `SIDEKIQ_ENABLED` is set we'll:
 The following ENV are available:
 
 - `SIDEKIQ_ENABLED`
+- `SIDEKIQ_QUEUES` - comma-separated custom queue names; if not specified, default queues are processed which are defined [here](./lib/roo_on_rails/sidekiq/settings.rb)
+- `SIDEKIQ_PERMITTED_LATENCY_VALUES` - our default queues come with a permitted latency value which is used in monitoring and Hirefire scaling. For your custom queues, you will need to specify the permitted latency in this format: `QUEUE_NAME:VALUE:UNIT` e.g. `queue-one:1:day,queue-two:3:minutes,queue-three:2:seconds`
```
gregbeech: It seems strange and error prone that there are two environment variables that need to match up for things to work, if they don't obey the `withinXunits` pattern. I'd suggest that we just use `SIDEKIQ_QUEUES` and have the latency as an optional setting, eg.

```
SIDEKIQ_QUEUES=within1hour,custom1:1day,within10minutes,custom2:3hours
```

i.e. the `:latency` is an optional part of the queue specification. I find the `1:day` syntax somewhat odd as it looks like a 3-tuple so I'd suggest either splitting on digits/non-digits as above, or something like `1.day` to make it clear it's a 2-tuple of `name:latency`.

lib/roo_on_rails/sidekiq/settings.rb
gregbeech: Also, you'll need to sort the queues by latency otherwise they'll be read in the wrong order. You _could_ rely on people always putting a 1hr queue before a 1day queue, but I probably wouldn't. If you _do_ rely on that, then you should at least document that lower latency queues must appear first.

spec/support/global_helpers.rb
```diff
@@ -0,0 +1,11 @@
+module GlobalHelpers
```
gregbeech: I'd be inclined to call this `EnvHelpers` or something like that, to make its purpose more clear.

spec/spec_helper.rb
```diff
@@ -11,6 +11,9 @@
 $LOAD_PATH.unshift File.expand_path('../..', __FILE__)
 
 RSpec.configure do |config|
+  require_relative './support/global_helpers.rb'
```
gregbeech: You should be able to use `require` here as the `$LOAD_PATH` is modified to include the `spec` folder (see line 11).

humzashah: i saw that, still it doesn't work. doesn't look like it looks into `spec/support`. plus, `require_relative` is O(1) which is why i prefer it until it's very ugly (file path is very long)

spec/support/global_helpers.rb
humzashah: agreed - done

README.md
humzashah: makes sense - done

lib/roo_on_rails/version.rb
humzashah: :+1: - done

lib/roo_on_rails/sidekiq/settings.rb
humzashah: I thought of that but doing it in one place would mean doing it in all places i.e. if we assume that a user can add a space after a comma, they might add a space anywhere. Still, I've appended a `strip` to wherever the user can add a space. And I've mentioned the sorting in the README now. - done

lib/roo_on_rails/sidekiq/queue_latency.rb
humzashah: makes sense - done

lib/roo_on_rails/sidekiq/queue_latency.rb
```diff
@@ -5,6 +5,29 @@ module Sidekiq
...
+            when 'default'
+              1.day.to_i
+            when /\Awithin\d+.+\z/
+              _, number, unit = queue_name.partition(/\d+/)
+              number.strip.to_i.public_send(unit.strip).to_i
+            else
+              ENV.fetch('SIDEKIQ_QUEUES', '').split(',').reduce(nil) do |result, entry|
+                entry = entry.strip
+                next result unless entry.start_with?(queue_name)
+                _, number, unit = entry.split(':').last.partition(/\d+/)
```
me: Since the parsing logic is reused, can we extract it?

lib/roo_on_rails/sidekiq/queue_latency.rb
```diff
@@ -5,6 +5,29 @@ module Sidekiq
...
+        @permitted_latency_values ||= Hash.new do |hash, queue_name|
+          hash[queue_name] = begin
+            case queue_name
+            when 'monitoring', 'realtime'
+              10.seconds.to_i
+            when 'default'
+              1.day.to_i
+            when /\Awithin\d+.+\z/
+              _, number, unit = queue_name.partition(/\d+/)
+              number.strip.to_i.public_send(unit.strip).to_i
```
me: Why the `strip`?

lib/roo_on_rails/sidekiq/queue_latency.rb
```diff
@@ -5,6 +5,29 @@ module Sidekiq
...
+            else
+              ENV.fetch('SIDEKIQ_QUEUES', '').split(',').reduce(nil) do |result, entry|
```
me: It would be cleaner have an external method that converts `SIDEKIQ_QUEUES` to a hash once I think

lib/roo_on_rails/sidekiq/queue_latency.rb
humzashah: https://github.com/deliveroo/roo_on_rails/pull/46#discussion_r128460495

me: I don't see how there could be a space in `number` though.

lib/roo_on_rails/sidekiq/queue_latency.rb
gregbeech: This loop here I found hard to understand. Maybe it would be simpler to build a `{name => latency}` hash of the `SIDEKIQ_QUEUES` variable once at startup and then this `else` block would just be a hash lookup?

lib/roo_on_rails/sidekiq/queue_latency.rb
```diff
@@ -5,6 +5,39 @@ module Sidekiq
     class QueueLatency
       extend Forwardable
 
+      DEFAULT_LATENCY_VALUES = {
```
me: I think this should live in `Sidekiq::Settings ` (it's weird that Settings has to ask `QueueLatency` for the list of queues).

humzashah: Yeah but with this approach we ensure that our queues are sorted by increasing permitted latency. Will keep it as is.

me: Why? What difference does it make for that where the constant lives?

humzashah: Are you suggesting that we simply move this constant to Settings?

me: Yes. Maybe renaming it to `DEFAULT_QUEUES`.

humzashah: Ok, I think I'll rename to DEFAULT_QUEUE_LATENCY_VALUES. `DEFAULT_QUEUES` doesn't give any context for the values in the Hash.

README.md
```diff
@@ -204,9 +205,18 @@ Run the following from your app's top-level directory:
...
-eventually.
+That command will sequentially run a number of checks. For it to run successfully, you will need:
+
+- a GitHub API token that can read your GitHub repository's settings placed in `~/.roo_on_rails/github-token`
+- the Heroku toolbelt installed and logged in
+- admin privileges on the `roo-dd-bridge-production` (this will be addressed eventually)
+- checks are run sequentially for staging and then for production. The process halts at any non-fixable failing check. To process only specific environments, you can set a config variable while running the command, like so:
+
+```
+ROO_ON_RAILS_ENVIRONMENTS=staging,production bundle exec roo_on_rails $(: this is the default behaviour)
```
mezis: I think the `$(: ... )` syntax might confuse non-bash-expert readers. Maybe just use `#`-comments?

lib/roo_on_rails/sidekiq/queue_latency.rb
```diff
@@ -1,4 +1,5 @@
 require 'active_support/core_ext/numeric'
+require_relative './settings'
```
mezis: We've been trying to _not_ use `require_relative` anywhere in this codebase, so far. Mind being explicit with the path?

spec/roo_on_rails/sidekiq/settings_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+          within5minutes
+          within30minutes
+          within1hour
+          within1day
+          default
+        ]
+      end
+    end
+
+    context 'custom value' do
```
mezis: I think we'd want tests for correctly-parsed latency values as well here.

README.md
humzashah: yup, done

lib/roo_on_rails/sidekiq/queue_latency.rb
humzashah: done

spec/roo_on_rails/sidekiq/settings_spec.rb
humzashah: ah, missed that; done

PR #47, opened 2017-07-21
Disable google auth by default

FriedSock:
The readme suggests that google auth is disabled by default, the code path for checking configuration will also default to disabled if the env var is not present; but the default env sets `GOOGLE_AUTH_ENABLED` to true.

If we actually do want to enable it by default, it is a major breaking change so we should issue a new major version of roo_on_rails.


mezis:
Unfortunately this now breaks the Omniauth tests.

PR #49, opened 2017-07-24
SETI-461: Convert OAuth to Railtie

mezis:
- Converts Google OAuth to Railtie, adds docs
- Tweaks integration test setup

===

Jira story [#SETI-461](https://deliveroo.atlassian.net/browse/SETI-461) in project *Software Engineering Tools and Infrastructure*:

> ## Why
> 
> Scaffolders/generators result in hard-to-maintain apps.
> 
> For instance, if the generator has a bug, and the generated code was modified, it becomes difficult to upgrade apps. Compare to fixing a bug in a Railtie and shipping a new gem version.
> 
> ## What
> 
> Convert from a generator to a Railtie.
> 
> Make a point to include the docs embedded in the current generated template in the README.


spec/support/build_test_app.rb
```diff
@@ -33,6 +33,7 @@ def ensure_scaffold
 
         shell_run "rails new #{scaffold_dir} #{rails_new_options}"
 
+        require 'rails'
```
FriedSock: 🎉 Does this mean that specs can be run in isolation now?

mezis: If you mean, "just run `bundle exec rspec` without `appraise`", then yes :)

PR #50, opened 2017-07-24
MetricsWorker reports only in-process queues

humzashah:
Previously, MetricsWorker was reporting metrics for all Sidekiq queues. This wasn't possible since the permitted latency of custom queues was defined through an environment variable which wasn't accessible in the Sidekiq process which MetricsWorker was running in.

This PR ensures that MetricsWorker only reports queues that are being processed by the current Sidekiq process.


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+<<<<<<< HEAD
+# v1.8.1 (2017-07-24)
```
mezis: You've left conflict marker in ;)
Also — can you specify the version as `HEAD` or `Unreleased`? The actual version number will be set by the maintainer (me, probably) when reviewing what was merged and deciding whether this needs to be a major or a minor.

humzashah: noooooooooooooooooo

spec/roo_on_rails/sidekiq/metrics_worker_spec.rb
```diff
@@ -5,148 +5,164 @@
...
-      allow(statsd).to receive(:batch).and_call_original
-      allow(statsd).to receive(:gauge)
+      allow(statsd).to receive(:gauge).with(any_args)
+      allow(Sidekiq::Queue).to receive(:all).and_return(sidekiq_queues) if defined?(sidekiq_queues)
     end
 
     after { reset_queues }
 
     it 'should send all metrics in a batch' do
+      expect(statsd).to receive(:batch).with(no_args).once
```
mezis: This is really odd to me - all the expectations have moved to _before_ we `perform` the actual task. I'm confused; why does it pass?

humzashah: rspec has 2 method-call expectations. `.to receive()` and `.to have_received()`. the former requires the call to be before the action is called. this way if we received some unexpected call, an error is raised. the former works but doesn't check for any 'wrong' calls.

spec/roo_on_rails/harness_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    let(:context_value) { {} }
+    let(:checks_class) { RooOnRails::Checks::Environment }
+    let(:staging_check) { instance_double(checks_class, run: true) }
+    let(:production_check) { instance_double(checks_class, run: true) }
+    let(:instance) {described_class.new(try_fix: try_fix, context: context_value) }
+
+    subject(:run) { instance.run }
+
+    before do
+      allow(checks_class).to receive(:new) do |options|
```
mezis: Stubbing `new` on classes (that are implementation detail) is a bit iffy... Any chance to revisit this a bit?
An idea could be, for instance, to create a "dry run" mode for checks, and have the harness return/expose the context so we can check what was attempted.

spec/roo_on_rails/sidekiq/metrics_worker_spec.rb
humzashah: i made this change because usually it's weird to have the execution of the code being tested in `before` block

mezis: Ah, duh. That's me being slow/tired. Ignore me.

CHANGELOG.md
humzashah: done, removed version from previous changelog entry as well

spec/roo_on_rails/harness_spec.rb
humzashah: done

PR #51, opened 2017-07-25
Don’t count quiet processes in the SLA current_process count

danielcooper:
Currently we use the full process count to calculate our desired count.

This change makes it so we don't use quiet sidekiqs that are due to be torn down. This stops spikes in the reported metric if queues stay around longer than expected on deployment


mezis:
Care to update the CHANGELOG?

lib/roo_on_rails/checks/environment.rb
```diff
@@ -6,7 +6,6 @@
 require 'roo_on_rails/checks/sidekiq/settings'
 require 'roo_on_rails/checks/heroku/drains_metrics'
 require 'roo_on_rails/checks/documentation/playbook'
-require 'roo_on_rails/checks/google_oauth/initializer'
```
gregbeech: Did you mean to remove this?

lib/roo_on_rails/sidekiq/process_scaling.rb
```diff
@@ -15,14 +15,15 @@ def initialize(queue_latencies)
 
       def current_processes
         ::Sidekiq::ProcessSet.new.count do |process|
+          process['quiet'] == 'false' &&
```
gregbeech: Eww this feels nasty... had a look at the Sidekiq code and it's a string because it's just the result of an `hmget` but I'd be inclined to either do `.to_s == 'false'` or convert it to a Boolean before comparing it just to ensure that if it does ever get 'fixed' to be a real Boolean this code continues to work.

lib/roo_on_rails/checks/environment.rb
mezis: Yeah. I'm trying to sheperd a bunch of PRs into master and ran into conflict hell. Fixing this as I write.

lib/roo_on_rails/sidekiq/process_scaling.rb
danielcooper: Seems fair to to_s. @mezis I don't want to step on your toes mid merging; shall I do this or will you?

PR #53, opened 2017-07-27
Speed up builds

mezis:
From this:

image

to this:

image


.travis.yml
```diff
@@ -13,16 +13,21 @@ gemfile:
...
 before_script:
   - unset RACK_ENV
   - unset RAILS_ENV
+  - mkdir -p $PWD/tmp && sudo mount -t tmpfs -o size=1024m tmpfs $PWD/tmp
 matrix:
   allow_failures:
     - rvm: ruby-head
-    - rvm: 2.4.0
+  exclude:
+    - rvm: 2.4.1
```
pedrocunha: Sorry i miss some context here but why exclude 2.4.1 in this section but above lists 2.4.1?

mezis: We're excluding just Rails 3 + Ruby 2.4.1 (which are incompatible), not all of 2.4.1.

PR #54, opened 2017-07-27
Google Oauth fix

mezis:
- Client apps would not load if they added any other middleware
- Using `RouteSet#prepend` happens to work in Rails 3 to 5.1

### DIFF
--- a/lib/roo_on_rails/railties/google_oauth.rb
+++ b/lib/roo_on_rails/railties/google_oauth.rb
@@ -1,4 +1,4 @@
-  require 'roo_on_rails/config'
+require 'roo_on_rails/config'
 
 module RooOnRails
   module Railties
--- a/lib/roo_on_rails/railties/google_oauth.rb
+++ b/lib/roo_on_rails/railties/google_oauth.rb
@@ -1,17 +1,13 @@
+  require 'roo_on_rails/config'
+
 module RooOnRails
   module Railties
     class GoogleOAuth < Rails::Railtie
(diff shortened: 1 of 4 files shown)


lib/roo_on_rails/railties/google_oauth.rb
```diff
@@ -1,36 +1,24 @@
...
-          # support development mode on route changes (only works in Rails 5)
-          app.reloader.to_prepare { _add_routes(app) }
         end
       end
 
       private
 
       def _if_enabled
         return unless Config.google_auth_enabled?
-        if Rails::VERSION::MAJOR < 5
```
knaveofdiamonds: Do you not still need this block?

mezis: No, because the new code actually works in Rails 3.2, 4.2, 5.0, and 5.1.

mezis:
> The changes look fine to me, but I don't understand what has changed here that fixes the middleware issue? Is it something around the disable_clear_and_finalize (even though that touches routes, not middleware), or just moving the require of roo_on_rails/config?

Appending our second initializer (the routes one) after `set_routes_reloader_hook` would cause all other initializers for the app (incl. adding further middleware) to also get loaded late, after the point where the middleware stack got committed and frozen.

So the "fix" is actually removing that 2nd initializer. That then required not fiddling with the reloader (which isn't available earlier), and making the routes still work — which is how I discovered `RouteSet#prepend`, which lets one declare a set of routes that get added to your app's.

PR #55, opened 2017-08-01
Allow app names to have environment abbreviations

halfwolf:
*Why* 
Heroku has a 30 character limit for app names. Requiring that the name includes  `-production` or `-staging` means that we have only 20 characters to name our services. 

*What*
Add `prd` and `stg` as acceptable environment names for `production` and `staging` respectively.


RenatoSerra22:
this would be awesome :) im currently restrained by this limit

lib/roo_on_rails/checks/heroku/app_exists.rb
```diff
@@ -16,6 +16,10 @@ module Heroku
       # Output context:
       # - heroku.app.{env}: an app name.
       class AppExists < EnvSpecific
+        ACCEPTABLE_ENV_NAMES = {
+          'production' => ['prod', 'production'],
+          'staging'    => ['stg', 'staging']
```
humzashah: did we reach any consensus on this? i'd suggest using 'test' or 'stag' instead of 'stg' so prod/test and prod/stag are both 4 chars long

humzashah: furthermore, do we want to ensure 'uniformity' - do we want to allow production to be named 'application-production' and staging to be named 'application-stag' or do we want to allow only valid sets:

- staging, production
- stag, prod
- test, prod

mezis: I guess whoever writes the code sets the convention—personally I'm fine either way.
I think it's not a huge deal as this will change when we change infra (DNS will be environment-local, so you won't need such suffixes).

mezis: I think we could perhaps use a mention of this in the README, tough?

halfwolf: Having done a little bit of research, there isn't much consensus around environment name abbreviations. We largely use `staging` so I think `test` would be a significant shift, and I have an emotional reaction to `stag` because said out loud it doesn't make me think of an environment 🦌 .  One source suggested `prd` and `stg` which I think aligns nicely with your point about character length @humzashah. 

My understanding of `roo_on_rails` is that it configures / validates each environment separately, so it'd be difficult to check that both environment use the same style of abbreviations, but I'm not very well versed in the code. @humzashah  @mezis  do you have any suggestions around how we could accomplish that?

humzashah: @halfwolf
- stg, staging, prd, and production work for me
- perhaps we can leave the 'uniformity' check - it's a nice-to-have but perhaps not super-important

PR #56, opened 2017-08-10
ISEC-94: Default and async Routemaster publishers

gregbeech:
While integrating the publishing library I found that:

* It was missing async publishing (and in my case linking availability
of the service to Routemaster by putting it on a hot path, which is
bad).

* I was repeating a bunch of boilerplate from the docs which could be
made into code that doesn’t need to be repeated.

As such, this adds async publishing as an option. This is at the
publisher level rather than in the model, because it’s possible you
might want multiple publishers for a model and one might be sync with
the other async.

It also adds the ability to register default publishers because most of
the time they’ll be the same and you can infer the URL in a common
publisher. I’ve changed the name to `ApplicationPublisher` in the docs
rather than `BasePublisher` to fit in with the general Rails naming
scheme.

### DIFF
--- a/README.routemaster_client.md
+++ b/README.routemaster_client.md
@@ -8,39 +8,29 @@ It also assumes that your app has an API for the resources you want to publish l
 
 ### Setup lifecycle events for your models
 
-We will most likely want to publish lifecycle events for several models, so to write slightly less code let's create a model concern first:
-
-```ruby
-# app/models/concerns/routemaster_lifecycle_events.rb
-require 'roo_on_rails/routemaster/lifecycle_events'
-
-module RoutemasterLifecycleEvents
-  extend ActiveSupport::Concern
-  include RooOnRails::Routemaster::LifecycleEvents
-
-  included do
-    publish_lifecycle_events
-  end
-end
-```
-
-Then let's include this concern to the relevant model(s):
+You can use publish events on create, update, and destroy by including the `PublishLifecycleEvents` module:
 
 ```ruby
 # app/models/order.rb
+require 'roo_on_rails/routemaster/publish_lifecycle_events'
+
 class Order < ApplicationRecord
-  include RoutemasterLifecycleEvents
+  include RooOnRails::Routemaster::PublishLifecycleEvents
 
   # ...
 end
 ```
 
-And another one for the example:
+If you need more control over which events are published you can use the base module `LifecycleEvents` and specify them explicitly:
 
 ```ruby
 # app/models/rider.rb
+require 'roo_on_rails/routemaster/publish_lifecycle_events'
+
 class Rider < ApplicationRecord
-  include RoutemasterLifecycleEvents
+  include RooOnRails::Routemaster::LifecycleEvents
+
+  publish_lifecycle_events :create, :destroy
 
   # ...
 end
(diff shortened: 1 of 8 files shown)


lib/roo_on_rails/routemaster/publisher.rb
```diff
@@ -22,7 +22,7 @@ def will_publish?
 
       def publish!
         return unless will_publish?
-        @client.send(@event, topic, url, data: stringify_keys(data))
+        @client.send(@event, topic, url, async: async?, data: stringify_keys(data))
```
danielcooper: Is this a feature of the client? What queue does it go into?

lib/roo_on_rails/routemaster/publisher.rb
```diff
@@ -33,6 +33,10 @@ def url
         raise NotImplementedError
       end
 
+      def async?
+        false
```
danielcooper: Maybe this should be true by default - hot path and all.

lib/roo_on_rails/routemaster/publisher.rb
gregbeech: https://github.com/deliveroo/routemaster-client#usage

lib/roo_on_rails/routemaster/publisher.rb
gregbeech: The default in `routemaster-client` is `false` so I thought I'd keep the same default otherwise it might be confusing.

mezis: Also, making the default `true` means you'd _have_ do have Sidekiq setup. I think we'd rather not force users to use all the features in `roo_on_rails`?

README.routemaster_client.md
```diff
@@ -8,66 +8,61 @@ It also assumes that your app has an API for the resources you want to publish l
...
+# app/publishers/application_publisher.rb
 require 'roo_on_rails/routemaster/publisher'
 
-class BasePublisher < RooOnRails::Routemaster::Publisher
+class ApplicationPublisher < RooOnRails::Routemaster::Publisher
   include Rails.application.routes.url_helpers
 
+  def url
+    url_helper = :"api_#{model.class.name.underscore}_url"
+    public_send(url_helper, model.id, host: ENV.fetch('API_HOST'), protocol: 'https')
```
kplattret: Now I'm thinking that the documentation's example should probably use UUIDs as it's what we've decided to go with and as it requires a [little trick](https://github.com/deliveroo/logistics-dashboard/pull/583/files#diff-38264488dcf6826516388149f59bbcbc) to work correctly.

README.routemaster_client.md
```diff
@@ -8,66 +8,61 @@ It also assumes that your app has an API for the resources you want to publish l
...
   # ...
 end
 ```
 
-And another one for the example:
+If you need more control over which events are published you can use the base module `LifecycleEvents` and specify them explicitly:
 
 ```ruby
 # app/models/rider.rb
+require 'roo_on_rails/routemaster/publish_lifecycle_events'
```
kplattret: You probably mean `roo_on_rails/routemaster/lifecycle_events`?

README.routemaster_client.md
gregbeech: I think it makes sense to leave the documentation relatively agnostic of how the IDs are created and/or what they are. Personally with UUID identifiers I would generate the identifier in the code before inserting the record, not in the database, then you wouldn't need to do anything like that.

PR #57, opened 2017-08-17
SETI-483: roo_on_rails aborts on the first failed check

mezis:
- Cleans up the master CLI, to support `roo_on_rails [sidekiq|harness]` subcommands
- Default to not fixing harness issue. Adds a `--fix` flag.
- Adds a `--env` flag to only run checks for a given environment (replaces the `ROO_ON_RAILS_ENVIRONMENTS` variable)
- Runs all checks even when one fails.

This last one is the masthead feature. Example run on `logistics-dashboard`:

image

...

image

(previously, this would have aborted at the first failed check)

===

Jira story [#SETI-483](https://deliveroo.atlassian.net/browse/SETI-483) in project *Software Engineering Tools and Infrastructure*:

> ## Why
> 
> The roo_on_rails command aborts on the first failed check, requiring some back-and-worth when multiple things need to be fixed.
> 
> ## What
> 
> Continue after a failed check, and mention when checks get skipped because dependencies are not met.

### DIFF
--- a/exe/roo_on_rails
+++ b/exe/roo_on_rails
@@ -70,7 +70,7 @@ module RooOnRails
             _options[:fix] = true
           end
 
-          o.on('-e', '--environment', 'Application environments to check') do |v|
+          o.on('-e', '--env', 'Application environments to check (comma-separated)') do |v|
             _options[:env] = v
           end
         end
--- a/exe/roo_on_rails
+++ b/exe/roo_on_rails
@@ -3,12 +3,104 @@
 require 'rubygems'
 require 'roo_on_rails/config'
 
-if ARGV.include? 'sidekiq'
-  require 'roo_on_rails/sidekiq/loader'
-  RooOnRails::Sidekiq::Loader.run
-else
-  require 'roo_on_rails/harness'
-  module RooOnRails
-    Harness.new(try_fix: true, context: Config.load).run
+
+module RooOnRails
+  module CLI
+    class Base
+      attr_reader :argv
+
+      def initialize(argv)
+        @argv = argv
+      end
+
+      def usage
+        raise NotImplementedError
+      end
+
+      protected
+
+      def _log(message)
+        $stderr.puts message
+      end
+
+      def _assert_argv(enum)
+        return if enum.include?(argv.length)
+        $stderr.puts enum.inspect
+        $stderr.puts argv.inspect
+        _log 'Wrong number of arguments.'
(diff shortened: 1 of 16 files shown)


exe/roo_on_rails
```diff
@@ -3,12 +3,103 @@
...
+        raise NotImplementedError
+      end
+
+      protected
+
+      def _log(message)
+        $stderr.puts message
+      end
+
+      def _assert_argv(enum)
```
gregbeech: This method is pretty confusing. It only ever seems to be called to assert that there are no _remaining_ arguments, and `enum` is actually a range (I see "enum" and I think enumeration type rather than range). Maybe `_assert_argv_length(range)` might be clearer, or `_assert_argv_empty` might be sufficient.

exe/roo_on_rails
```diff
@@ -3,12 +3,103 @@
...
+        require 'optionparser'
+
+        OptionParser.new do |o|
+          o.banner = usage
+
+          o.on('-f', '--fix', 'Attempt to fix failed checks') do
+            _options[:fix] = true
+          end
+
+          o.on('-e', '--environment', 'Application environments to check') do |v|
```
gregbeech: Should this be `--env` according to the README? Also it's not clear to me how this would work if I wanted to check multiple environments? If I did `--env staging --env production` then this would only check `production` because it would overwrite `_options[:env]` the second time?

mezis: It's passed a comma-separated list (although typically you'd use it with just one item, since `staging,production` is the default).

Will clarify and fix `--env` v. `--environment`, thank you.

lib/roo_on_rails/checks/base.rb
```diff
@@ -84,10 +96,15 @@ def signature
...
-        end
+        context.deps ||= {}
+        deps.map { |dep|
+          sig = dep.signature.join('/')
+          if context.deps.key?(sig)
+            context.deps[sig]
+          else
+            context.deps[sig] = dep.run
+          end
+        }.to_a.all?
```
gregbeech: Isn't `.to_a` superfluous here? Not a huge deal.

mezis: I _think_ that in Ruby 2.4 it's required, as `all?` is lazy and we want to evaluate all items (not stop at the first failure).

gregbeech: Ah, good point. Sorry!

PR #60, opened 2017-08-25
SETI-477: Uniform logging

mezis:
Changes:

- Deprecated `RooOnRails::ContextLogging`
- Replaces the Rails loggers with a compatible structured logger (mostly equivalent to `ContextLogging`)
- Lowers logging noise from Railties 

===

Jira story [#SETI-477](https://deliveroo.atlassian.net/browse/SETI-477) in project *Software Engineering Tools and Infrastructure*:

> ## Why
> 
> Uniform and structured logging facilitates several activities — from diagnosis to post-facto investigations.
> We want to standardise this.
> 
> ## What
> 
> Provide an overriden `Rails.logger` that meets the following criteria:
> 
> - Output should be valid `logfmt`
> - Timestamps should only be emitted when the output is a TTY (ie. in development mode)
> - Log level should be normalised to e.g. `at=INFO`
> - The message text should be `msg="..."`
> - There should always be a context specified as the `in=...` field. Depending on context this should be the request ID (for web threads), the job ID (for sidekiq jobs), or as a fallback the thread ID.
> -- It should be possible to override this context.
> - Extra logfmt fields can be passed to the logger.
> 
> Sample invocation:
> ```
> Rails.logger.info('hello world', foo: 'bar')
> # at=INFO in=6d284dd2-6d94-4283-8782-4618ea433a53 foo="bar" msg="hello world" {code} 
> ```
> NB: datastore-monitor and Routemaster have more-or-less sane implementations of this.


lib/roo_on_rails/logger.rb
```diff
@@ -0,0 +1,92 @@
...
+      _context_stack.replace([{}])
+    end
+
+    def set_log_level
+      self.level = ::Logger::Severity.const_get(ENV.fetch('LOG_LEVEL', 'DEBUG'))
+    end
+
+    private
+
+    class Proxy < SimpleDelegator
```
marcinwyszynski: Nit: maybe move the inner class before the `private` block, since it isn't really private? Or maybe you want make it private. Either way, the intention here is a bit unclear.

lib/roo_on_rails/logger.rb
```diff
@@ -0,0 +1,92 @@
...
+      def initialize(logger, context)
+        @context = context
+        super(logger)
+      end
+
+      %w[add debug info warn error fatal unknown].each do |name|
```
marcinwyszynski: Do you think that small savings in avoided boilerplate really warrant those arcane incantations?

spec/roo_on_rails/logger_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+    context 'when the log method is chained within an overlapping context block' do
+      before { logger.with(i: 123, s: 'stuff') { logger.with(s: 'nonsense').info('hello-world') } }
+
+      it 'should output the message then the flattened context as logfmt' do
+        expect(output).to eq 'at=INFO msg=hello-world i=123 s=nonsense'
+      end
+    end
+  end
+
+  # describe '#flush' do
```
marcinwyszynski: Please remove commented out code.

lib/roo_on_rails/railties/logging.rb
```diff
@@ -0,0 +1,16 @@
...
+  module Railties
+    class Logging < Rails::Railtie
+      initializer 'roo_on_rails.logging.before', before: :initialize_logger do
+        require 'roo_on_rails/logger'
+        Rails.logger = config.logger = RooOnRails::Logger.new
+        Rails.logger.debug 'initializer roo_on_rails.logging.before'
+      end
+
+      initializer 'roo_on_rails.logging.after', after: :initialize_logger do
+        Rails.logger.set_log_level
```
tompave-roo: isn't this called inside `RooOnRails::Logger#initialize` already? Is this here because the `LOG_LEVEL` ENV var could change in the meantime? Well, it's doesn't hurt doing it again I think, this is just a question :-)

lib/roo_on_rails/logger.rb
mezis: It's intended to be private, hence placing it here — event though as you point out Ruby doesn't have private subclasses.

lib/roo_on_rails/logger.rb
mezis: Er, I'm not sure what's arcane here? The usage of `define_method`? This is actually _less_ arcane than [the way ActiveSupport does it](https://github.com/rails/rails/blob/master/activesupport/lib/active_support/logger.rb#L92-L98), in my humble opinion.

lib/roo_on_rails/railties/logging.rb
mezis: It is, unfortunately Rails resets the log level based on what's in `config/environments/{development,production}.rb` — and those _will_ contain a loglevel in a fresh Rails apps. Which we want to override :)

lib/roo_on_rails/logger.rb
marcinwyszynski: It kind of does - you can use `private_constant`.

lib/roo_on_rails/logger.rb
marcinwyszynski: That, in conjunction with `__getobj__` and `public_send`. And it saves you maybe 12 lines of boilerplate. I'm not going to block on that or anything, but I hate metaprogramming, and IMO this is one of those cases where the benefit is not worth the cost.

README.md
```diff
@@ -149,38 +149,47 @@ be used:
...
+```
+at=INFO msg="hello world"
 ```
 
-You can then add context using the `with` method:
+One can also add context using the `with` method:
 
 ```ruby
 logger.with(a: 1, b: 2) { logger.info 'Stuff' }
+# => at=INFO msg=Stuff a=1 b=2
```
gregbeech: Not quite sure I understand the `at` name here? Wouldn't `level` be more obvious?

lib/roo_on_rails/logfmt.rb
```diff
@@ -26,18 +22,10 @@ def dump_value(v)
...
-
-        escaped = str.gsub(/(["\\])/, '\\\\\1')
-        %("#{escaped}")
-      end
 
-      def ident?(str)
-        str.bytes.all? { |b| b > SPACE && b != EQUALS && b != QUOTE }
+        @_escape_re ||= /[[:space:]"']/
+        return str unless @_escape_re =~ str
+        str.inspect
```
gregbeech: Nice simplification 👍

lib/roo_on_rails/logger.rb
```diff
@@ -0,0 +1,104 @@
...
+  #
+  #   logger.with(a: 1, b: 2).info('Stuff')
+  #   # Logs "at=INFO msg=Stuff a=1 b=2"
+  #
+  #   logger.with(a: 1) { logger.with(b: 2).info('Stuff')  }
+  #   # Logs "at=INFO msg=Stuff a=1 b=2"
+  #
+  # Hashes, arrays and any complex object that supports `#to_json` will be
+  # output in escaped JSON format so that it can be parsed out of the attribute
+  # values.
```
gregbeech: According to the readme you can also do `logger.info('Stuff', a: 1, b: 2)` so might be worth adding that in here if you want to be comprehensive.

lib/roo_on_rails/railties/database.rb
```diff
@@ -3,7 +3,7 @@ module Railties
     class Database < Rails::Railtie
       initializer 'roo_on_rails.database', after: 'active_record.initialize_database' do
         ActiveSupport.on_load :active_record do
-          $stderr.puts 'initializer roo_on_rails.database'
+          Rails.logger.debug 'initializer roo_on_rails.database'
```
gregbeech: Might be nice to change these to something like `initializer: 'roo_on_rails.database'` now we can ;-)

lib/roo_on_rails/railties/logging.rb
gregbeech: Yeah the whole logging setup in rails is a bit of a mess.

lib/roo_on_rails/railties/sidekiq.rb
```diff
@@ -11,18 +11,18 @@ class Sidekiq < Rails::Railtie
...
         else
-          $stderr.puts 'skipping initializer roo_on_rails.sidekiq'
+          Rails.logger.debug 'skipping initializer roo_on_rails.sidekiq'
         end
       end
 
       def config_hirefire(app)
         unless ENV['HIREFIRE_TOKEN']
-          warn 'No HIREFIRE_TOKEN token set, auto scaling not enabled'
+          Rails.logger.warn 'No HIREFIRE_TOKEN token set, auto scaling not enabled'
```
gregbeech: Might be nice to include `initializer: 'roo_on_rails.sidekiq` as part of the metadata for these messages; you could use a `with` block around them to do it.

README.md
mezis: I've been using `at=` for consistency with what Heroku does, and what [the original `logfmt` paper](https://brandur.org/logfmt) says.

lib/roo_on_rails/logger.rb
mezis: According to the README... but not in practice (the current code does not support it). Do you think this is a feature we should add (back)?

lib/roo_on_rails/railties/database.rb
mezis: Ooh, good idea. The `msg` is still mandatory though, guess I'll have it say just `"loading"`.

lib/roo_on_rails/logger.rb
```diff
@@ -0,0 +1,104 @@
...
+      if @show_timestamp
+        "[%s] %7s | %s %s\n" % [
+          datetime.utc.strftime(TIMESTAMP_FORMAT),
+          severity,
+          message,
+          Logfmt.dump(_context_stack.last)
+        ]
+      else
+        "%s\n" % Logfmt.dump({
+          at:   severity,
```
danielcooper: Is that supposed to be `at`? Looks like a time field.

mezis: See comment above —
> I've been using `at=` for consistency with what Heroku does, and what [the original `logfmt` paper](https://brandur.org/logfmt) says.

lib/roo_on_rails/railties/sidekiq.rb
mezis: Done - this call is wrapped inside a `Logger#with` block now.

lib/roo_on_rails/logger.rb
mezis: Actually, I can't see what you mention in the README :/
Happy to introduce this in a further PR though.

PR #61, opened 2017-08-29
SETI-479: Move roo_on_rails to CircleCI

mezis:
Moves from Travis to CircleCI for reasonably fast builds.

Pros:

- No queueing (unlike Travis for OSS)
- Seems less flaky (fewer timeout failures)

Cons:

- Individual jobs are 2x slower. E.g. for Rails 5 on Ruby 2.3.4, Travis took [171 seconds](https://travis-ci.org/deliveroo/roo_on_rails/jobs/270081633) and Circle took [326 seconds](https://circleci.com/gh/deliveroo/roo_on_rails/175), in spite of similar conditions (fully cached dependencies, RAM disk for scaffolded tests apps).
- Configuration is more complex (and generated through ERb).

===

Jira story [#SETI-479](https://deliveroo.atlassian.net/browse/SETI-479) in project *Software Engineering Tools and Infrastructure*:

> ## Why
> 
> Every build of roo_on_rails takes 30+ minutes, because it's open-source and cannot benefit from the parallelism on our paid account.
> 
> ## What
> 
> Either figure a way to build on the paid Travis account, or move to CircleCI (now that they have build matrix support)


.circleci/config.yml.erb
```diff
@@ -0,0 +1,84 @@
...
+
+      - run: 
+          name: Install dependencies
+          command: |
+            bundle install --jobs=3 --retry=3 --path=$PWD/vendor/bundle
+
+      - run:
+          name: Run test suite
+          command: |
+            unset RACK_ENV &&
```
marcinwyszynski: Missing backslashes?

.circleci/config.yml.erb.orig
```diff
@@ -0,0 +1,90 @@
...
+          name: Setup RAM disk
+          command: |
+            rm -rf ~/project/tmp ;
+            mkdir /dev/shm/tmp ;
+            ln -s /dev/shm/tmp ~/project/tmp
+
+      - run:
+          name: Select build variant (Ruby <%= ruby %>, <%= variant %>)
+          command: |
+<<<<<<< HEAD
```
marcinwyszynski: Ugh.

.circleci/config.yml.erb
mezis: Not necessary thanks to the pipe above.

marcinwyszynski: My `irb` interpreter says otherwise:

```
irb(main):001:0> require 'yaml'
=> true
irb(main):002:0> YAML
=> Psych
irb(main):003:0> YAML.load(File.read('hello.yaml'))
=> {"hello"=>"one\ntwo\nthree\n"}
irb(main):004:0> 
marcinw@bruno yaml
$ cat hello.yaml 
hello: |
  one
  two
  three
```

I think what you'd want to use in this case is `>`:

```
irb(main):013:0> require 'yaml'
=> true
irb(main):014:0> YAML.load(File.read('hello.yaml'))
=> {"hello"=>"one two three\n"}
irb(main):015:0> 
marcinw@bruno yaml
$ cat hello.yaml 
hello: >
  one
  two
  three
```

marcinwyszynski: I guess it works because Bash treats dangling `&&` as expecting further input:

```
marcinw@bruno yaml
$ unset HELLO &&
> unset BOOM &&
> bundle exec rspec
```

PR #62, opened 2017-08-31
Use model names not classes to avoid issues

jphastings:
I was having some trouble with Roo On Rails using an incorrect publisher, demonstrated with this pseudo-irb session, so I've created a PR here to use the names of the model classes, rather than the classes. I've not investigated why the objects would be different, but it's likely because of a `reload!` in a rails console, but I feel like this library should be able to cope with that.

```irb
> MyModel
=> MyModel(id: integer, updated_at: datetime)

> pubs = RooOnRails::Routemaster::Publishers.instance_variable_get('@publishers')
=> {MyModel(id: integer, updated_at: datetime)=>#<Set: {MyModelPublisher}>}

> RooOnRails::Routemaster::Publishers.for(MyModel.first, :update)
=> [#<ApplicationPublisher:0x007fc220718790 @_routes=nil, @model=#<MyModel id: 243, updated_at: "2017-08-30 19:50:03">, @event=:update, @client=Routemaster::Client>]

> pubs[MyModel]
=> nil

> MyModel
=> MyModel(id: integer, updated_at: datetime)

> pubs.keys.first
=> MyModel(id: integer, updated_at: datetime)

> MyModel.object_id
=> 70235881143460

> pubs.keys.first.object_id
=> 70235880994880

```

### DIFF
--- a/lib/roo_on_rails/routemaster/publishers.rb
+++ b/lib/roo_on_rails/routemaster/publishers.rb
@@ -9,12 +9,12 @@ module RooOnRails
       end
 
       def self.register(publisher_class, model_class:)
-        @publishers[model_class] ||= Set.new
-        @publishers[model_class] << publisher_class
+        @publishers[model_class.name] ||= Set.new
+        @publishers[model_class.name] << publisher_class
       end
 
       def self.for(model, event)
-        publisher_classes = @publishers[model.class] || @default_publishers
+        publisher_classes = @publishers[model.class.name] || @default_publishers
         publisher_classes.map { |c| c.new(model, event) }
       end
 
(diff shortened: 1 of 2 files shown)


lib/roo_on_rails/routemaster/publishers.rb
```diff
@@ -5,16 +5,16 @@ module Publishers
       @publishers = {}
 
       def self.register_default(publisher_class)
-        @default_publishers << publisher_class
+        @default_publishers << publisher_class.name
```
gregbeech: Oops, actually here you want the class not the name of the publisher.

Unless you're going to store the publishers as names, and then do `constantize.new`

PR #64, opened 2017-09-01
Declare the event's timestamp according to model

jphastings:
When the model responds to `#created_at` or `#updated_at` then make an intelligent guess as to the timestamp the event should be declared as having.

Also introduces a `#timestamp` method on a publisher which can override this.


mezis:
Looks excellent. Can we haz a `CHANGELOG` entry?

jphastings:
Your wish is my command ☺️

jphastings:
As an aside @mezis, could we add a requirement with codeclimate that there's _always_ a change to `CHANGELOG.md` for a PR to be accepted?

PR #66, opened 2017-09-08
Fixed an issue with papertrail `fix` command.

martyphee:
There was a mismatch between the method definition and how it was being called.


jphastings:
You're probably going to need to get some tests in there for this to be mergable @martyphee but otherwise get it up to master and we'll ship :)

jphastings:
Also, @martyphee generally you should leave changes to the version number out of PRs, so the release can be done separate to features being added :)

martyphee:
I put a test in to make sure the call to `fix` won't fail again.

PR #67, opened 2017-09-21
Add a controller concern for authenticating routes

jphastings:
This concern allows API authentication in a consistent manner.

If a service connects with basic auth using the username "service" then the
`SERVICE_CLIENT_KEY` environment variable must have the given password as one of
the comma separated strings within it or a 403 will be raised.

Any service with an acceptable key can access routes:
```ruby
class ThingController < ApplicationController
  include RooOnRails::Concerns::RequireApiKey
  require_api_key

  # etc
end
```

Only the specified clients can access specific routes in this controller:
```ruby
class ThingController < ApplicationController
  include RooOnRails::Concerns::RequireApiKey
  require_api_key(only_services: :my_service, only: :create)

  # etc
end
```

### DIFF
--- a/lib/roo_on_rails/concerns/require_api_key.rb
+++ b/lib/roo_on_rails/concerns/require_api_key.rb
@@ -40,8 +40,8 @@ module RooOnRails
         # @see AbstractController::Callbacks::ClassMethods#before_action for additional scoping opts
         def require_api_key(only_services: nil, **options)
           before_action(**options) do
-            authenticate_or_request_with_http_basic('actor-tracking') do |service_name, client_key|
-              Authenticator.new(*only_services).valid?(service_name, client_key).tap do |is_valid|
+            authenticate_or_request_with_http_basic('Authenitcation required') do |service_name, client_key|
+              Authenticator.new([*only_services]).valid?(service_name, client_key).tap do |is_valid|
                 @current_client = OpenStruct.new(name: service_name).freeze if is_valid
               end
             end
--- a/lib/roo_on_rails/concerns/require_api_key.rb
+++ b/lib/roo_on_rails/concerns/require_api_key.rb
@@ -1,0 +1,108 @@
+require 'active_support/concern'
+require 'action_controller/metal/http_authentication'
+
+module RooOnRails
+  module Concerns
+    # This concern allows API authentication in a consistent manner.
+    #
+    # If a service connects with basic auth using the username "service" then the
+    # `SERVICE_CLIENT_KEY` environment variable must have the given password as one of
+    # the comma separated strings within it or a 403 will be raised.
+    #
+    # @example: Any service with an acceptable key can access routes
+    #
+    #     class ThingController < ApplicationController
+    #       include RooOnRails::Concerns::RequireApiKey
+    #       require_api_key
+    #
+    #       # etc
+    #     end
+    #
+    # @example: Only the specified clients can access specific routes in this controller
+    #
+    #     class ThingController < ApplicationController
+    #       include RooOnRails::Concerns::RequireApiKey
+    #       require_api_key(only_services: :my_service, only: :create)
+    #
+    #       # etc
+    #     end
+    module RequireApiKey
+      extend ActiveSupport::Concern
+      include ActionController::HttpAuthentication::Basic::ControllerMethods
+
+      attr_reader :current_client
+
+      module ClassMethods
+        # Declares that routes on the controller must have access credentials specified
+        # in the request that match the approparite environment variables.
+        #
+        # @param :only_services (#to_s,Array<#to_s>) Restricts the services which will be accepted
+        # @see AbstractController::Callbacks::ClassMethods#before_action for additional scoping opts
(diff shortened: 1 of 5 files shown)


lib/roo_on_rails/concerns/require_api_key.rb
```diff
@@ -0,0 +1,108 @@
...
+
+      module ClassMethods
+        # Declares that routes on the controller must have access credentials specified
+        # in the request that match the approparite environment variables.
+        #
+        # @param :only_services (#to_s,Array<#to_s>) Restricts the services which will be accepted
+        # @see AbstractController::Callbacks::ClassMethods#before_action for additional scoping opts
+        def require_api_key(only_services: nil, **options)
+          before_action(**options) do
+            authenticate_or_request_with_http_basic('actor-tracking') do |service_name, client_key|
```
martyphee: Should `actor-tracking` be here?

lib/roo_on_rails/concerns/require_api_key.rb
```diff
@@ -0,0 +1,108 @@
...
+              Authenticator.new(*only_services).valid?(service_name, client_key).tap do |is_valid|
```
martyphee: If `only_services` is nil will this cause an exception?

lib/roo_on_rails/concerns/require_api_key.rb
jphastings: It should _not_! Good spot!

lib/roo_on_rails/concerns/require_api_key.rb
jphastings: `*nil` -> `[]`

jphastings: Good spot!

lib/roo_on_rails/concerns/require_api_key.rb
```diff
@@ -0,0 +1,108 @@
...
+
+      module ClassMethods
+        # Declares that routes on the controller must have access credentials specified
+        # in the request that match the approparite environment variables.
+        #
+        # @param :only_services (#to_s,Array<#to_s>) Restricts the services which will be accepted
+        # @see AbstractController::Callbacks::ClassMethods#before_action for additional scoping opts
+        def require_api_key(only_services: nil, **options)
+          before_action(**options) do
+            authenticate_or_request_with_http_basic('Authenitcation required') do |service_name, client_key|
```
gregbeech: "Authentication"

lib/roo_on_rails/concerns/require_api_key.rb
```diff
@@ -0,0 +1,108 @@
...
+              end
+            end
+          end
+        end
+      end
+
+      # This functionality pulled out into a new class for testability
+      class Authenticator
+        def initialize(whitelisted_clients)
+          @whitelisted_clients = whitelisted_clients.map(&:to_s)
```
gregbeech: I'd probably make this a `Set` for O(1) lookup rather than O(n).

lib/roo_on_rails/concerns/require_api_key.rb
```diff
@@ -0,0 +1,108 @@
...
+          return false unless whitelisted?(service_name)
+
+          NewRelic::Agent.add_custom_attributes(httpBasicUserId: service_name) if defined?(NewRelic)
+          ClientApiKeys.instance.valid?(service_name, client_key)
+        end
+
+        private
+
+        def whitelisted?(service_name)
+          return true if @whitelisted_clients.empty?
```
gregbeech: This line isn't really needed; `include?` should be O(1) if it's empty anyway.

lib/roo_on_rails/concerns/require_api_key.rb
```diff
@@ -0,0 +1,108 @@
...
+          @cache = ENV.select { |key| key =~ CLIENT_KEY_NAME_SUFFIX_REGEX }
+                      .map { |k, v| [service_name(k), parse_client_keys(v)] }
+                      .to_h
+                      .freeze
+        end
+
+        def valid?(service_name, client_key)
+          return false if service_name == '' || client_key == ''
+
+          client_keys = @cache[normalize(service_name)]
```
gregbeech: I realise I wrote this initially, but I'd probably have a `INVALID_KEYS = Set.new(SecureRandom.hex(20))` (i.e. a set with one key) and then change this line to `@cache[normalize(service_name)] || INVALID_KEYS` then remove the `client_keys &&` from the subsequent line, so we're always doing a lookup in a set. This should reduce the possibility for any timing attack to discover valid service names.

lib/roo_on_rails/concerns/require_api_key.rb
jphastings: D'oh!

jphastings:
Thanks for the feedback @gregbeech; I've got to move on to other work now, but I'll be happy to pick this up these changes in the future if you'd like, or feel free to make alterations yourself!

PR #74, opened 2017-10-18
Fix: Handle the case when service_name is nill in auth

mvalipour:
### Background

It appears, due to bad configuration, `service_name` may be nil in the authentication logic. This needs to be handled gracefully so the system will continue with an `unauthorised` rather than an unhandled error [showing up in NewRelic](https://rpm.newrelic.com/accounts/881102/applications/53286913/traced_errors/1df4d3d5-b3ea-11e7-beb1-0242ac110008_0_12325)

### DIFF
--- a/lib/roo_on_rails/concerns/require_api_key.rb
+++ b/lib/roo_on_rails/concerns/require_api_key.rb
@@ -83,7 +83,7 @@ module RooOnRails
         end
 
         def valid?(service_name, client_key)
-          return false if service_name == '' || client_key == ''
+          return false if service_name.to_s.empty? || client_key.to_s.empty?
 
           client_keys = @cache[normalize(service_name)]
           return false unless client_keys
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,10 @@
-# v1.12.0 (2017-10-12)
+# v1.13.1 (2017-10-18)
+
+Bug fix:
+
+- Fixes issue when `service_name` is null in `require_api_key` (#74)
+
+# v1.13.0 (2017-10-12)
 
 Features:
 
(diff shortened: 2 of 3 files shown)


lib/roo_on_rails/concerns/require_api_key.rb
```diff
@@ -83,7 +83,7 @@ def initialize
         end
 
         def valid?(service_name, client_key)
-          return false if service_name == '' || client_key == ''
+          return false unless service_name && client_key
```
Crunch09: What do you think about `if service_name.to_s == '' || client_key.to_s == ''` ? `unless` with `&&` is always harder to follow for me 😵

mvalipour: I'd personally avoid the cost of additional `.to_s` -- and quite frankly find `unless` more readable 😄 

But don't feel strong about it, if that's the general preference, I'm happy to change

jphastings: If we're worried about the cost of `to_s` then we should be coding in Assembler, not in Ruby 😉 

What you've written so far doesn't actually cover both cases though; we need to duplicate `blank?` functionality — it needs to return `false` if either are nil or empty string (which is also covered in Florian's comment).

Maybe: `if service_name.to_s.empty? || client_key.to_s.empty?`?

mvalipour: I understand the edge case -- I don't see a test case covering it tho! 😉

jphastings:
This looks good @mvalipour, could you add a line to the CHANGELOG.md then we can merge when green and I'll cut a release :)

lib/roo_on_rails/concerns/require_api_key.rb
mvalipour: BTW, in this case we don't need to catch the whitespace here, given it will then continue and fail finding a valid key for `''` -- but I made the change anyway :)

PR #78, opened 2017-11-02
LOG-2010: Roo on Rails: fix sending events t to Routemaster

me:
Publisher was sending epoch seconds instead of milliseconds as RM expects.

===

Jira story [#LOG-2010](https://deliveroo.atlassian.net/browse/LOG-2010) in project *Logistics*:

> Roo on Rails is senting "t" in seconds, but Routemaster expects milliiseconds. Fix it.

### DIFF
--- a/lib/roo_on_rails/version.rb
+++ b/lib/roo_on_rails/version.rb
@@ -1,3 +1,3 @@
 module RooOnRails
-  VERSION = '1.13.1'.freeze
+  VERSION = '1.14.0'.freeze
 end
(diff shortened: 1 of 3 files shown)


lib/roo_on_rails/version.rb
```diff
@@ -1,3 +1,3 @@
 module RooOnRails
-  VERSION = '1.13.1'.freeze
+  VERSION = '1.14.0'.freeze
```
danielcooper: Best to do the version updates in a releasing PR

me: Makes sense. Do you think for now I can merge like it is or should I split it?

danielcooper: I'd split it, but I'm not well versed in releasing.

(You'll also wanna do a changelog entry)

PR #79, opened 2017-11-08
Middleware for accepting JWTs

jphastings:
- Adds a `PopulateEnvFromJWT` middleware for Rack, loadable with `require 'roo_on_rails/railties/roo_identity'`
- Checks the Authorization header for a JWT and populates `roo.identity` in the rack environment with the claims, if present and valid.

GRO-1245


lib/roo_on_rails/rack/populate_env_from_jwt.rb
```diff
@@ -0,0 +1,98 @@
...
+require 'faraday'
+require 'faraday_middleware'
+
+module RooOnRails
+  module Rack
+    class PopulateEnvFromJWT
+      UnnacceptableKeyError = Class.new(RuntimeError)
+      VALID_JWK_URL_PREFIXES = [
+        'https://www.deliveroo.com/identity-keys/',
+        'https://identity.deliveroo.com/identity-keys/'
```
gregbeech: Identity is on `identity.deliveroo.net` at the moment. Also, you'll need some way to make this configurable for staging so that `identity-staging.deliveroo.net` will work in that environment.

lib/roo_on_rails/rack/populate_env_from_jwt.rb
```diff
@@ -0,0 +1,98 @@
...
+      ].freeze
+
+      def initialize(app, logger:)
+        @app = app
+        @keys = {}
+        @logger = logger
+
+        if development?
+          @logger.warn "JWTs will not have their signatures checked, you're in development mode."
```
gregbeech: This seems odd... wouldn't we want to check signatures even in development mode?

lib/roo_on_rails/rack/populate_env_from_jwt.rb
```diff
@@ -0,0 +1,98 @@
...
+        env['roo.identity'] = decode_authorization_header(env['HTTP_AUTHORIZATION'])
+        @app.call(env)
+
+      # Other exceptions will bubble up, allowing the higher middleware to return a 500, which is
+      # intentional.
+      rescue UnnacceptableKeyError, JSON::JWT::Exception => e
+        # Identifying user is clearly attempting to hack or has been given a totally incorrect
+        # token, log this and flag as Forbidden, without executing the rest of the middleware stack.
+        ::NewRelic::Agent.notice_error(e) if defined?(NewRelic)
+        [403, {}, []]
```
gregbeech: This should really be a `401`:

* `401 Unauthorized` - I either don't know who you are or don't _believe_ you are who you say you are.
* `403 Forbidden` - I know and believe who you are, but you cannot do what you're trying to do.

lib/roo_on_rails/rack/populate_env_from_jwt.rb
```diff
@@ -0,0 +1,98 @@
...
+
+        ENV['RACK_ENV'] == 'development'
+      end
+
+      # @raise [UnnacceptableKeyError,Faraday::Error,OpenSSL::OpenSSLError] From `#public_key`
+      # @raise [JSON::JWT::Exception] Bubble ups from `JSON::JWT.decode`
+      # @return [JSON::JWT] The list of claims this header makes by way of a JWS token. Will be an
+      #   empty hash for invalid or absent tokens.
+      def decode_authorization_header(header_value)
+        return JSON::JWT.new unless (header_value || '').starts_with?('Bearer ')
```
gregbeech: Can you use `header_value&.starts_with?('Bearer ')` here, or are we using too old a version of Ruby for RoR?

lib/roo_on_rails/rack/populate_env_from_jwt.rb
```diff
@@ -0,0 +1,98 @@
...
+      rescue Faraday::ParsingError
+        raise JSON::JWT::InvalidFormat, 'Downloaded JWK is not a valid JSON file'
+      end
+
+      def http_request
+        Faraday.new do |conf|
+          conf.response :json
+          conf.response :raise_error
+          conf.request :json
+          conf.use FaradayMiddleware::FollowRedirects, limit: 3
```
gregbeech: I'd probably be inclined _not_ to follow redirects for something as critical as a key; that's an additional attack vector that somebody could try to exploit. I'd say if it's not at the exact URL directly then failing would be the best option.

lib/roo_on_rails/rack/populate_env_from_jwt.rb
gregbeech: Also, this may have further-reaching concerns than you expect. If the page the user wants to visit doesn't require authentication then this probably shouldn't fail just because the credentials they're presenting are invalid. It could reasonably just ignore them.

lib/roo_on_rails/rack/populate_env_from_jwt.rb
jphastings: That's a good point. The only reason I added it was because `www.deliveroo.com` redirects to `deliveroo.co.uk`, but indeed it's not a good vulnerability to open!

lib/roo_on_rails/rack/populate_env_from_jwt.rb
jphastings: Roo On Rails supports old rubies (for now) :(

lib/roo_on_rails/rack/populate_env_from_jwt.rb
jphastings: Ahh; for some reason I thought 403 was "no matter what you do, you can't do this" (because you're trying to do nefarious things with your dodgy token). I'll make the change :)

lib/roo_on_rails/rack/populate_env_from_jwt.rb
jphastings: Then you'd have to have a running orderweb/identity in order to run your service (plus we'd have to add `https://identity.deliveroo.com.dev` as a prefix, and point out that this service _needs_ to be at that location for the service you're developing to run).

This also means that there's a much more limited risk of putting an insecure key in the codebase to make development easier and having that key be left in for production 😱

lib/roo_on_rails/rack/populate_env_from_jwt.rb
jphastings: Hmm, yeah, I was thinking about environments (see the `.dev` stuff below). I'll add per-environment hash here, which isn't ideal, but it'll get the job done.

lib/roo_on_rails/rack/populate_env_from_jwt.rb
jphastings: To your second comment: I considered just passing "no claims" through to the rest of the middleware stack; but the only cases where this block should be executed is if the token _could never_ have been correct (ie. maliciously crafted, or something has been incorrectly configured) — which is why it pings NewRelic and gives up.

lib/roo_on_rails/rack/populate_env_from_jwt.rb
gregbeech: Fair enough. Maybe want to make it optional that you can check sigs in dev though in case you want to test it? Even if the default is not to?

lib/roo_on_rails/rack/populate_env_from_jwt.rb
gregbeech: The key difference is that with a 401 reauthenticating might help (which here it might; maybe somehow your token got corrupted and you could reauthenticate to get a new one). With a 403 reauthenticating will never help as the system already knows who you are and you're not allowed.

lib/roo_on_rails/rack/populate_env_from_jwt.rb
```diff
@@ -0,0 +1,98 @@
+require 'json/jwt'
+require 'faraday'
+require 'faraday_middleware'
+
+module RooOnRails
+  module Rack
+    class PopulateEnvFromJWT
+      UnnacceptableKeyError = Class.new(RuntimeError)
```
mcmillan: Is there a typo here? Looks like there's an extra `n` in `Unacceptable`.

lib/roo_on_rails/rack/valid_identity_service_prefixes.yml
```diff
@@ -0,0 +1,6 @@
+production:
+  - https://deliveroo.co.uk/identity-keys/
+  - https://identity.deliveroo.net/identity/keys/
+staging:
+  - https://test.deliveroo.co.uk/identity-keys/
+  - https://identity-staging.deliveroo.net/identity-keys/
```
Crunch09: Can this list be overwritten? Especially because `roo_on_rails` is opensource people might want to overwrite it in their application?

jphastings: They could, but they can't overwrite it in our services, which is the only place that matters to us :)

spec/roo_on_rails/rack/populate_env_from_jwt_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+
+      it { should be_ok }
+      include_examples 'roo.identity provided to inner app'
+
+      context 'when the signature is right but the specified algorithm is incorrect' do
+        # There is an attack vector where the nefarious individual uses the _public_ key to
+        # sign the JWT, but specifies the HMAC signing mechanism. A poorly secured client will
+        # see the HMAC declaration in the header, load the public key _as a string_ and check
+        # the signature as if the public key was a shared password in an HMAC signature.
+        # This test ensures that is not possible :)
```
gregbeech: 👍

PR #80, opened 2017-11-16
Add ability to create exceptions to SSL enforcement

mcmillan:
As we move more services onto AWS, we need to serve a health check
over HTTP that returns an 200 status code. At the moment, this
doesn't work for apps using `roo_on_rails` – `Rack::SslEnforcer`
redirects the health check request, which means it returns a 301
status code, and the health check fails.

This adds an environment variable – the ever-so-catchily named
`ROO_ON_RAILS_SSL_ENFORCEMENT_EXCEPTIONS` – that allows developers
to specify a comma-delimited list of exact paths that should be
excluded from SSL enforcement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,14 @@ We'll insert the following middlewares into the rails stack:
    with `ROO_ON_RAILS_RACK_DEFLATE` (default: 'YES').
 4. Optional middlewares for Google Oauth2 (more below).
 
+#### Adding an exception to `Rack::SslEnforcer`
+
+If you're running your application on Hopper and need to serve a health check
+over HTTP, or if for any other reason you might need to serve a particular endpoint
+over HTTP rather than HTTPS, you can use the `ROO_ON_RAILS_SSL_ENFORCEMENT_EXCEPTIONS`
+environment variable to specify a comma-delimited list of endpoints, for example
+`ROO_ON_RAILS_SSL_ENFORCEMENT_EXCEPTIONS=/health`.
+
 ### Database configuration
 
 The database statement timeout will be set to a low value by default. Use
--- a/lib/roo_on_rails/railties/http.rb
+++ b/lib/roo_on_rails/railties/http.rb
@@ -18,7 +18,7 @@ module RooOnRails
             ::Rack::Timeout
           )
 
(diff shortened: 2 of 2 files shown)


gregbeech:
Obvious question: Why not change Hopper to make the request over HTTPS?

mcmillan:
@gregbeech The request is made by an AWS ALB (which is performing SSL termination) directly to the EC2 machines, so at this point we have no concept of SSL being available or otherwise – it's all HTTP all the way down until the ALB handles SSL termination for us.

gregbeech:
Less obvious point: This probably won't work anyway.

Hopper runs outside the new infrastructure as it's part of the control plane, and thus it'll be talking to the applications using their public endpoints. The public endpoints will be locked down so that they can only be accessed via Cloudflare for security reasons (i.e. to prevent cloud piercing attacks). Cloudflare rewrites all HTTP URLs to HTTPS before it gets anywhere near your application. So it'll get a 301 at the edge anyway.

gregbeech:
> The request is made by an AWS ALB

Ah, ignore my point above then.

mcmillan:
@gregbeech Apologies – should have made that more clear from the offset.

@Crunch09 I've been thinking of a way to effectively test this and it seems rather difficult. Our [existing middleware tests](https://github.com/deliveroo/roo_on_rails/blob/925a1da64f4af5218163fb0c1bf4e1209b7daae8/spec/integration/http_spec.rb) involve invoking `rake middleware` which doesn't give us the ability to validate the options passed to a piece of middleware, just that the given piece of middleware exists. Do you have any ideas?

Crunch09:
@mcmillan Ah, didn't notice the `unless Rails.env.test?` in the middleware which is probably there for a good reason, so all fine by me 👍

mcmillan:
So – it turns out that the middleware actually needs to be removed entirely in order for apps to function on Hopper infrastructure (as we don't know the state of SSL in app instances) – HTTP-to-HTTPS redirects are performed at edge level in Cloudflare. I'm closing this PR and will replace it with a new one later.

jphastings:
Can we get a `CHANGELOG.md` addition?

PR #81, opened 2017-11-16
Ensure JWTs work across environments

jphastings:
Using `RACK_ENV` is not a valid way of distinguishing between production and staging, as for both of those environments it will be "production"!

This moves to using the `VALID_IDENTITY_URL_PREFIXES` environment variable (and will crash non-development environments which don't have it set) so that you must always set the places where keys are valid from.

### DIFF
--- a/spec/roo_on_rails/rack/populate_env_from_jwt_spec.rb
+++ b/spec/roo_on_rails/rack/populate_env_from_jwt_spec.rb
@@ -8,6 +8,7 @@ describe RooOnRails::Rack::PopulateEnvFromJWT, :webmock do
   let(:env) { { 'HTTP_AUTHORIZATION' => auth_header } }
 
   let(:rack_env_var) { 'production' }
+  let(:identity_url_prefixes) { 'https://deliveroo.co.uk/identity-keys/' }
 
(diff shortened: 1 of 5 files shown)


spec/roo_on_rails/rack/populate_env_from_jwt_spec.rb
```diff
@@ -123,12 +119,20 @@
...
+        let(:identity_url_prefixes) { 'https://another.deliveroo.tld/identity-keys/,https://identity.deliveroo.net/jwks/' }
+        let(:jku) { 'https://identity.deliveroo.net/jwks/0.jwk' }
 
-      its(:status) { should eq 401 }
+        it { should be_ok }
+        include_examples 'roo.identity provided to inner app'
+      end
+
+      context 'when the jku specified is not whitelisted' do
+        let(:jku) { 'https://hax0rs.com/sadface.jwk' }
```
gregbeech: 😄

gregbeech:
Seems reasonable. The other option (which is roughly equally nasty) is to use something like `SECURITY_ENV=staging` which is what we did in the identity service for its YAML config.

PR #82, opened 2017-11-16
Add ability to disable SSL enforcement

mcmillan:
This adds an option to RoR to disable SSL enforcement, which is
required for applications deployed to Hopper.

### DIFF
--- a/lib/roo_on_rails/railties/http.rb
+++ b/lib/roo_on_rails/railties/http.rb
@@ -18,7 +18,7 @@ module RooOnRails
             ::Rack::Timeout
           )
 
-          middleware_to_insert_before = Rails::VERSION::MAJOR < 4 ? ::ActionDispatch::Cookies : ::Rack::Head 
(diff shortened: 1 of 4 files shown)


lib/roo_on_rails/railties/http.rb
```diff
@@ -34,7 +34,7 @@ class HTTP < Rails::Railtie
           end
 
           # Don't use SslEnforcer in test environment as it breaks Capybara
-          unless Rails.env.test?
+          unless Rails.env.test? || ENV.fetch('ROO_ON_RAILS_DISABLE_SSL_ENFORCEMENT', 'YES').to_s =~ /\A(YES|TRUE|ON|1)\Z/i
```
mbranch: I really don't have any context, but I'd worry that setting this var to `YES` by default could be a bad idea—any downstream users of this gem would have SSL turned off, maybe unexpectedly.

mcmillan: D'oh – you're totally right – that default `YES` is in there erroneously. Now fixed.

mcmillan:
@jphastings Ready to rock and roll whenever you are – have rebased.

PR #86, opened 2018-01-04
[LOG-2207] Roo on Rails: allow publishing of events even for unchanged models

Natalienka:
We need a way to re-emit created/updated/deleted events later on (when the model is not a new record anymore and it does not have any changes on it).

===
Jira story [#LOG-2207](https://deliveroo.atlassian.net/browse/LOG-2207) in project Logistics

### DIFF
--- a/lib/roo_on_rails/routemaster/publisher.rb
+++ b/lib/roo_on_rails/routemaster/publisher.rb
@@ -4,26 +4,24 @@ require 'routemaster/client'
 module RooOnRails
   module Routemaster
     class Publisher
-      attr_reader :model, :event, :force_publish
+      attr_reader :model, :event
 
-      def initialize(model, event, client: ::Routemaster::Client, force_publish: false)
+      def initialize(model, event, client: ::Routemaster::Client)
         @model = model
         @event = event
         @client = client
-        @force_publish = force_publish
       end
 
       def publish?
-        return true if force_publish
         noop? || @model.new_record? || @model.previous_changes.any?
       end
 
-      def will_publish?
-        Config.routemaster_publishing_enabled? && publish?
+      def will_publish?(force_publish: false)
+        Config.routemaster_publishing_enabled? && (force_publish || publish?)
       end
 
-      def publish!
-        return unless will_publish?
+      def publish!(force_publish: false)
+        return unless will_publish?(force_publish: force_publish)
         @client.send(
           @event,
           topic,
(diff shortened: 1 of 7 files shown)


lib/roo_on_rails/routemaster/publisher.rb
```diff
@@ -4,15 +4,17 @@
...
 
-      def initialize(model, event, client: ::Routemaster::Client)
+      def initialize(model, event, client: ::Routemaster::Client, force_publish: false)
         @model = model
         @event = event
         @client = client
+        @force_publish = force_publish
       end
 
       def publish?
```
jphastings: Is there a reason you have this as a class instance variable, rather than an argument to `publish?`?

I'd imagine that `publisher.publish?(force: true)` would read better than having to scan back to wherever the publisher is initialized to know whether or not it'll be published or not.

Natalienka:
We don't actually mean to send the event multiple times. We've recently come across an incident where we thought that no events were published from dispatcher for some time. Fortunately it turned out not to be true, but we want to be ready for this kind of issue and we need to be able to emit events for objects that were created/updated during some period of time in the past.

jphastings:
That sounds like something worthwhile preparing for! 😅 Why would we need to force sending `create` and `update` though? Would a `noop` not be enough to catch the system back up?

lib/roo_on_rails/routemaster/publisher.rb
Natalienka: I just felt it better fits the current code since the `publish?` method is called through `will_publish?` method so it will have to receive `force_publish` parameter too. But `will_publish?` method is actually not forced to publish (it checks in config whether publishing is enabled). But if you think it makes more sense that way, I can definitely change it.

Natalienka:
I don't really know to be honest. I've already seen services that ignored `noop` events completely.

Natalienka:
@jphastings According to discussion on Slack in #event-bus-fu it's not a good idea to send these events as `noop` since most of the apps wouldn't handle them.

I've reconsidered your comment and changed `force_publish` to be an argument passed to `publish!` method. I'm not sure if it would make sense to also create another version of `publish_lifecycle_events` method that would force publishing (this was suggested on Slack).

Natalienka:
Good point @Crunch09 , I've updated the `CHANGELOG.md` file. Can I please get the approval again?

PR #87, opened 2018-01-09
Respect rails config log level

FriedSock:
The current behaviour is: when LOG_LEVEL is not set, then default the 
log level to DEBUG; regardless of what is set in
`config/environments/production.rb` for example.


lib/roo_on_rails/railties/logging.rb
```diff
@@ -8,7 +8,8 @@ class Logging < Rails::Railtie
       end
 
       initializer 'roo_on_rails.logging.after', after: :initialize_logger do
-        Rails.logger.set_log_level
+        log_level = Rails.configuration.log_level
+        Rails.logger.set_log_level(default: log_level)
```
mcmillan: If `log_level` is nil, what is the behaviour here? Will `set_log_level` attempt to call `upcase` on it and the app will blow up?

FriedSock: Rails will raise an error if `log_level` is set to something invalid (and it defaults to debug) so it cannot end up being nil at app boot. But yes, there should be a test here to check that the validation of this doesn't occur later on in the boot sequence.

FriedSock: I checked, and the setter itself will raise an error if you set it to something invalid. I've added a comment to explain this as I don't want to write a test for Rails behaviour.

FriedSock:
@jphastings 👍 done, I also had to add a commit due to `pg` publishing version 1.0.0 yesterday which broke the test suite 🎉

PR #88, opened 2018-01-11
Change New Relic initialization to use init_plugin

me:
Currently, app startup gives an ugly warning on New Relic initialization. This happens because the `manual_start` seems to be buggy. 

However, the New Relic behaviour on Rails is to call `init_plugin` (which `manual_start` eventually calls too). I think we should follow what New Relic does in its initialiser.

Additionally, the default behaviour in Rails is to set `sync_startup` to false, while `manual_start` sets it to true. I don't want to change the default behaviour here, as it could have unexpected consequences (very short lived processes would potentially not be able to send reports). However, we should allow apps to run with `sync_startup` set to false, in case startup time is important and the process is long-living. This allows setting the `NEW_RELIC_SYNC_STARTUP` environment variable (as per New Relic config), defaulting it to `true`.

### DIFF
--- a/lib/roo_on_rails/railties/new_relic.rb
+++ b/lib/roo_on_rails/railties/new_relic.rb
@@ -19,7 +19,7 @@ module RooOnRails
             abort "Aborting: newrelic.yml detected in '#{path.parent.realpath}', should not exist"
           end
 
-          sync_startup = ENV.fetch('NEW_RELIC_SYNC_STARTUP', 'YES').match?(/\A(YES|TRUE|ON|1)\Z/i)
+          sync_startup = (ENV.fetch('NEW_RELIC_SYNC_STARTUP', 'YES') =~ /\A(YES|TRUE|ON|1)\Z/i)
 
           require 'newrelic_rpm'
           unless Rails.env.test?
--- a/lib/roo_on_rails/railties/new_relic.rb
+++ b/lib/roo_on_rails/railties/new_relic.rb
@@ -19,7 +19,7 @@ module RooOnRails
             abort "Aborting: newrelic.yml detected in '#{path.parent.realpath}', should not exist"
           end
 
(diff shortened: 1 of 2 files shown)


lib/roo_on_rails/railties/new_relic.rb
```diff
@@ -19,8 +19,12 @@ class NewRelic < Rails::Railtie
             abort "Aborting: newrelic.yml detected in '#{path.parent.realpath}', should not exist"
           end
 
+          sync_startup = !(ENV['NEW_RELIC_SYNC_STARTUP'] == 'false')
```
humzashah: `ENV['NEW_RELIC_SYNC_STARTUP'].to_s.downcase != 'false'`

me:
There seems to be some serious flakiness with CI...

lib/roo_on_rails/railties/new_relic.rb
```diff
@@ -19,8 +19,12 @@ class NewRelic < Rails::Railtie
             abort "Aborting: newrelic.yml detected in '#{path.parent.realpath}', should not exist"
           end
 
+          sync_startup = !(ENV['NEW_RELIC_SYNC_STARTUP'].to_s.downcase == 'false')
```
danielcooper: I'm a fan of using this pattern for these

```
ENV.fetch('NEW_RELIC_SYNC_STARTUP', 'YES').match?(/\A(YES|TRUE|ON|1)\Z/i)
```

pedrocunha: To be honest, if only we could just say it's either 0s or 1s that would make our lifes much better

humzashah: I second what Pedro said for the future. 0/1 or false/true. Handling them is much easier and cleaner:

```ruby
ENV[key].to_i == 1
ENV[key].to_s.downcase == 'true'
```

me:
@danielcooper can you merge this? thanks!

PR #91, opened 2018-04-03
Updates Sidekiq Pro statsd integration syntax

nikz:
Newer versions of `sidekiq-pro` no longer take options when creating middleware. This change uses a Proc to set the dogstatsd instance as is the current style.

### DIFF
--- a/lib/roo_on_rails/railties/sidekiq.rb
+++ b/lib/roo_on_rails/railties/sidekiq.rb
@@ -39,8 +39,16 @@ module RooOnRails
       end
 
       def config_sidekiq_metrics
+        # https://github.com/mperham/sidekiq/wiki/Pro-Metrics
         require 'sidekiq-pro'
         ::Sidekiq::Pro.dogstatsd = -> { RooOnRails.statsd }
+
+        Sidekiq.configure_server do |config|
+          config.server_middleware do |chain|
+            require 'sidekiq/middleware/server/statsd'
+            chain.add Sidekiq::Middleware::Server::Statsd
+          end
+        end
       rescue LoadError
         Rails.logger.warn 'Sidekiq metrics unavailable without Sidekiq Pro'
       end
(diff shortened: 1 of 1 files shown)


lib/roo_on_rails/railties/sidekiq.rb
```diff
@@ -40,13 +39,8 @@ def config_sidekiq
       end
 
       def config_sidekiq_metrics
-        require 'sidekiq/middleware/server/statsd'
-
-        ::Sidekiq.configure_server do |x|
-          x.server_middleware do |chain|
-            chain.add ::Sidekiq::Middleware::Server::Statsd, client: RooOnRails.statsd
```
alsutton: I think you still need the chain.add

alsutton:
Merging because code climates shouldn't block progress.

nikz:
@pedrocunha can we check for the existence of the constant, or would you like us to add a config setting?

pedrocunha:
I think considering this already behing a SIDEKIQ_ENABLED we should just check the existence of the constant. Maybe ideally rolling back to the previous person if is still possible

What do you think ?

On 4 May 2018, 17:17 +0800, Nik Wakelin <notifications@github.com>, wrote:
> @pedrocunha can we check for the existence of the constant, or would you like us to add a config setting?
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.

-- 




This email, including any attachments, is confidential and may be 
privileged. If you are not the intended recipient please notify the sender 
immediately, and please delete it; you should not copy it or use it for any 
purpose or disclose its contents to any other person. If the email, 
including any attachments, is legally privileged, you should not forward it 
to any other person.

nikz:
@pedrocunha it does `rescue` the `LoadError` below.

This is an API change from Sidekiq, so we could check the version if necessary to support older versions. Or print a warning to upgrade?

PR #92, opened 2018-04-11
[next] Update versions and drop old ones

tompave-roo:


roo_on_rails.gemspec
```diff
@@ -23,7 +23,7 @@ Gem::Specification.new do |spec|
 
   spec.add_runtime_dependency 'dotenv-rails', '~> 2.1'
   spec.add_runtime_dependency 'newrelic_rpm'
-  spec.add_runtime_dependency 'rails', '>= 3.2.22', '< 5.2'
+  spec.add_runtime_dependency 'rails', '>= 3.2.22', '< 5.3'
```
danielcooper: Should be >= 4.0?

tompave-roo: Ops! Good catch.

PR #93, opened 2018-04-11
[next] Drop HireFire

tompave-roo:


danielcooper:
What's the plan for future scaling? Can't see a way to do SLA based without some app co-operation.

tompave-roo:
@danielcooper That's why I left Sidekiq's SLA module in the repo. Will still be useful in the future.

PR #98, opened 2018-04-11
[next] StatsD tags

tompave-roo:


lib/roo_on_rails/statsd.rb
```diff
@@ -33,19 +34,41 @@ def env_name
       ENV['STATSD_ENV'] || ENV['HOPPER_ECS_CLUSTER_NAME'] || 'unknown'
     end
 
+    # Identifies a specific source container by type (e.g. web, worker)
+    # and unique UUID.
+    #
     def source_name
-      ENV['DYNO'] || ENV['HOSTNAME'] || 'unknown'
+      container_name = ENV.fetch('HOPPER_ECS_CONTAINER_NAME', 'unknown')
```
danielcooper: We've run into trouble in the past with extra tags (or tags that change often) messing up some aggregations in data dog.

Can we have an ENV to switch this off?

tompave-roo: I added this because it's closer to what we have on heroku, that is it takes the form `web.X` or `worker.X`, where `X` is an integer on Heroku and a UUID on ECS.

> trouble in the past with extra tags (or tags that change often)

If there are dashboard that are actively using the `source` tags, and the need continuity, I can see why this would be a problem. Yes. making this opt-in would make sense.

Since this aims to be a major version bump, from `1.21.x` to `2.0`, is it a deal breaker?

tompave-roo: What I'm saying is that the upgrade is not automatic now mandatory, and apps can still customize the source tag if they need it: https://github.com/deliveroo/rs-admin/blob/staging/server/config/initializers/00_statsd.rb#L11

danielcooper: Yeah, I suppose an app can override. 

Much hassle to wrap in an env anyway? Happy for it to be on by default.

tompave-roo: Sure, I can do it 👍

PR #101, opened 2018-04-16
Support rails 5.2

novotnyjakub:
Apart from adding build support for Rails `5.2.x` it also updates the build versions of Ruby to the latest patches available in the `deliveroo/multiruby` Docker image.


Appraisals
```diff
@@ -18,7 +18,14 @@ appraise 'rails-5' do
 end
 
 appraise 'rails-5-1' do
-  gem 'rails', '~> 5.1'
+  gem 'rails', '~> 5.1.0'
+  gem 'sqlite3'
```
pedrocunha: why does it need the news gems?

novotnyjakub: Well I made sure that this appraisal is only for `5.1.x` and not for `5.x`. The `sqlite` I'm not sure, to be honest, but it's in all of the other appraisal groups as well. I think the tests were failing without it.

pedrocunha:
This looks. How do you feel about adding 2.5 since we are here?

novotnyjakub:
Btw. @pedrocunha I can't merge this, because I don't have permission. So please feel free to merge this PR so that we can release it. Whenever you have some time, of course. Thanks!

PR #105, opened 2018-08-29
LOG-3076: Report to New Relic per Hopper service.

mjkillough:
If `BASE_NEW_RELIC_APP_NAME` is set, report to New Relic for each Hopper
service individually *and* report everything to one central New Relic
application. If `NEW_RELIC_APP_NAME` is specified, report to one New
Relic application as before.

The Reliability Monitor requires that each web/worker service reports to
New Relic separately in order to get statistics about request
counts/errors.

### DIFF
--- a/lib/roo_on_rails/railties/new_relic.rb
+++ b/lib/roo_on_rails/railties/new_relic.rb
@@ -12,6 +12,15 @@ module RooOnRails
 
           abort 'Aborting: NEW_RELIC_LICENSE_KEY is required' if license_key.nil?
 
+          # Report application stats to a per-service (worker, web) New Relic app, and to a main
+          # application for all services.
+          base_name = ENV['BASE_NEW_RELIC_APP_NAME']
+          service_name = ENV['HOPPER_SERVICE_NAME']
+          if !base_name.blank? && ENV['NEW_RELIC_APP_NAME'].blank?
+            task_app_name = service_name.present? ? "#{base_name} - #{service_name}" : nil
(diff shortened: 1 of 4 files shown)


lib/roo_on_rails/railties/new_relic.rb
```diff
@@ -12,6 +12,15 @@ class NewRelic < Rails::Railtie
 
           abort 'Aborting: NEW_RELIC_LICENSE_KEY is required' if license_key.nil?
 
+          # Report application stats to a per-service (worker, web) New Relic app, and to a main
+          # application for all services.
+          base_name = ENV['BASE_NEW_RELIC_APP_NAME']
+          service_name = ENV['HOPPER_SERVICE_NAME']
+          if !base_name.nil? && ENV['NEW_RELIC_APP_NAME'].blank?
```
me: Should these `nil?` checks be `blank?`?

lib/roo_on_rails/railties/new_relic.rb
```diff
@@ -12,6 +12,15 @@ class NewRelic < Rails::Railtie
...
+            task_app_name = !service_name.nil? ? "#{base_name} - #{service_name}" : nil
```
me: Might be more readable with `service_name.present?`

lib/roo_on_rails/railties/new_relic.rb
mjkillough: I think you're right.

lib/roo_on_rails/railties/new_relic.rb
mjkillough: Ahh, yes! :-) Much nicer.

PR #113, opened 2019-05-21
SRE-xxx: Fix broken CI

VSpike:
The master branch no longer passes CI because of various problems.

* Release of Bundler v2
* Release of `sqlite` gem version 1.4.x
* Adding in Ruby 2.5.x causes tests to fail

These changes switch to using the latest Ruby 2.4.x and 2.5.x versions, and also breaks the link to the deliveroo/multiruby base container.


exe/roo_on_rails
```diff
@@ -36,6 +36,11 @@ module RooOnRails
       end
 
       def run
+        # Force output to be unbuffered, otherwise the CI
+        # tests will fail because the parent app won't see
+        # the log output from the test app
+        $stdout.sync = true
```
danielcooper: Do you need both this and the line in the run_test_app?

VSpike: Yes, I think so! When it tests sidekiq, it starts the `exe/roo_on_rails` entrypoint. For other tests, it calls `bin/rails`.

VSpike:
> Do we have plans to support Ruby 2.6.X at all?

Don't see why not!  A few upstream projects are using it so it would make sense.

PR #114, opened 2019-05-23
Remove New Relic

btslater:
Given that we are moving off New Relic at the end of the month, remove the New Relic gem and all the code associated with it.


lib/roo_on_rails/routemaster/lifecycle_events.rb
```diff
@@ -31,7 +30,8 @@ def publish_event(event, force_publish:)
           begin
             publisher.publish!(force_publish: force_publish)
           rescue => e
-            NewRelic::Agent.notice_error(e)
+            # SENTRY: To be migrated once we're on sentry
+            Rails.logger.error("#{e.inspect} rescued in #{self.class.name}::publish_event")
```
rxbynerd: this unfortunately won't get surfaced anywhere meaningful — can we do something like `Raven.report_exception(e) if defined?(Raven)` or something, to make sure we're not losing these?

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# v2.0.0
```
gregbeech: Two questions:

- Did you update the gem version? I can't see a version change in the code itself?
- Given we probably want to remove a bunch of other Heroku/Hirefile related stuff (see "NEXT" open PR) would it be worth using something like `2.0.0.pre1` so we can do a bunch of pre-releases before getting to the final 2.0.0 with everything gone? Otherwise each removal will be another major version, then we'll be up to a much bigger number very soon.

lib/roo_on_rails/rack/populate_env_from_jwt.rb
```diff
@@ -41,7 +41,6 @@ def call(env)
       rescue UnacceptableKeyError, JSON::JWT::Exception => e
         # Identifying user is clearly attempting to hack or has been given a totally incorrect
         # token, log this and flag as Forbidden, without executing the rest of the middleware stack.
-        ::NewRelic::Agent.notice_error(e) if defined?(NewRelic)
```
farthir: Would it be worth logging this with Rails.logger and Raven as below?

lib/roo_on_rails/routemaster/lifecycle_events.rb
```diff
@@ -31,7 +30,7 @@ def publish_event(event, force_publish:)
           begin
             publisher.publish!(force_publish: force_publish)
           rescue => e
-            NewRelic::Agent.notice_error(e)
+            Raven.report_exception(e) if defined?(Raven)
```
farthir: Might also be worth logging with `Rails.logger` so it's still recorded in Datadog for services without Sentry?

lib/roo_on_rails/rack/populate_env_from_jwt.rb
btslater: Worth logging with Raven - but Rails isn't in scope here so we can't use logger

PR #119, opened 2020-05-21
[DEFF-565] replace calls to logger.with with just logger

mariogintili:
# DEFF-565

Replaces `Rails.logger.with` with just `Rails.logger`. To address 2 problems:

1) Some rails applications hosting this library won't have `RooOnRails::Logger` as its default logger. 

2) Injecting `RooOnRails::Logger` as the default logger does not guarantee that other libraries inside of a project may inject a different logger which implements a different API than `RooOnRails::Logger`


.circleci/config.yml
```diff
@@ -178,9 +194,17 @@ jobs:
   build_2.4.6_rails_5_1:
     docker:
       - image: ruby:2.4.6-alpine
+        environment:
```
joeyj-deliveroo, line 197: you could use a reusable block to cut down the copy/paste :) or if you update the version to 2.1, you can use commands: https://circleci.com/docs/2.0/reusing-config/

reusable block:
```yaml
common_ruby_environment: &common_ruby_environment
  environment:
    RAILS_ENV: test
    PGHOST: 127.0.0.1
    PGUSER: postgres
    POSTGRES_HOST_AUTH_METHOD: "trust"
common_pg_environment_vars: &common_pg_environment_vars
    POSTGRES_USER: postgres
    POSTGRES_DB: ror_testapp_test
    POSTGRES_HOST_AUTH_METHOD: "trust"
.
.
.
  build_2.4.6_rails_5_1:
    docker:	    docker:
      - image: ruby:2.4.6-alpine
       <<: *common_ruby_environment
      - image: postgres:9.6-alpine
         environment:
          PGDATA: /dev/shm/pgdata
          <<: *common_pg_environment_vars
```

mariogintili:
@joeyj-deliveroo It's versioned so it wont affect any services other versions,

the test suite uses the library inside of a rails app

PR #121, opened 2020-09-17
Rails 6 compatibility

pedrocunha:
What it says on the tin


VSpike (approved):
Awesome! Thanks.

spec/support/build_test_app.rb
```diff
@@ -36,9 +36,14 @@ def ensure_scaffold
         scaffold_dir.rmtree if scaffold_dir.exist?
         TEST_DIR.mkpath
 
-        shell_run "rails new #{scaffold_dir} #{rails_new_options}"
-
         require 'rails'
+
+        if Rails::VERSION::MAJOR > 5
+          shell_run "rails new #{scaffold_dir} #{rails_new_options} --skip-javascript"
```
alexstoick, line 42: why did we add `--skip-javascript`?

lib/roo_on_rails/version.rb
```diff
@@ -1,3 +1,3 @@
 module RooOnRails
-  VERSION = '2.0.2'.freeze
+  VERSION = '2.1.0'.freeze
```
alexstoick, line 2: minor bump for major support for a rails version 🤔

alexstoick (approved):

spec/support/build_test_app.rb
pedrocunha: Good question. We either do this or we have to require `webpacker` as a gem dependency for rails 6. Since we aren't going to be building new rails apps with frontends in rails I think this is fine as coverage.

lib/roo_on_rails/version.rb
pedrocunha: I think this is OK since the gem is still backwards compatible and hasn't been following Rails versions I am afraid.

spec/support/build_test_app.rb
alexstoick: Fair enough ✅

alexstoick (approved):

PR #124, opened 2021-05-28
Fix env from jwt bug

byarr:
keys and mapped_urls were sharing the same dictionary.

During normal use this doesn't seem to break anything. But if the first call to get a key fails - then we will cache the mapped url in the keys dictionary. Then try and use this URL as a key object, this fails.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# v2.1.2
+
+Bug fix:
+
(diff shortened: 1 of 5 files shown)


egisj (approved):
🐛

jphastings (dismissed):

jphastings:
Good spot! Thanks bud 😊

egisj (approved):

PR #129, opened 2022-06-06
Update Sidekiq

DanielVartanov:
This PR removes the Sidekiq version constraint. It is meant to be used by teams to upgrade their Sidekiq version to 6+


sljory (approved):
lgtm

